PR #7, opened 2019-04-02
LT-908: I want to view the details of a specific case in a work queue

jfaracik:
- Open a case and view details about it


jfaracik:
> Preferable to separate out the LT-908 features and the addition of the npm pack. Can this be done?

More so removing files than adding them. I'm happy to keep it in here considering it's a minimal change.

jfaracik:
> Please remove the changes for the npm installation from this PR

Resolved with team. Please approve.

PR #22, opened 2019-05-11
feat: LT-911 - Add notes to cases

jfaracik:
- Added case notes to case screen
- You can now add a case note to a case
- Validation for said case note creation
- Using the new Forms error page generator for showing errors to the user


assets/javascripts/specific/case_notes.js
```diff
@@ -0,0 +1,49 @@
...
+
+$("#case_note").on('input propertychange paste', function() {
+	if ($(this).val() != '') {
+		$(".lite-expandable-textarea-controls").removeClass("lite-expandable-textarea-controls--hidden");
+	} else {
+		$(".lite-expandable-textarea-controls").addClass("lite-expandable-textarea-controls--hidden");
+	}
+
+	if ($(this).val().length > 1200) {
+		$("#case_note-warning").text("You have " + (2200 - $(this).val().length) + " character" + pluralize(2200 - $(this).val().length) + " remaining");
```
lukesa: use a variable ```maximum_lenght``` instead of 2200

jfaracik: Added three constants for min_length, warning visible_length, max_length

PR #25, opened 2019-05-31
Lt 930 add departments

lukesa:


lukesa:
> The lite-api needed to be updated - Fixed now
> 
> The story talks about a Save and a Cancel button. Th UI only has a Submit button. Would that do?

Form class doesn't support cancel button yet. Use back button instead

PR #31, opened 2019-06-17
LT-936 adding Gov Users and managing them

cstobart:


auth/views.py
```diff
@@ -44,6 +45,9 @@ def get(self, request, *args, **kwargs):
                 client_secret=settings.AUTHBROKER_CLIENT_SECRET,
                 code=auth_code)
 
+            print('BANANA')
```
jfaracik: Banana

users/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
+
```
jfaracik: Empty?

users/models.py
```diff
@@ -0,0 +1,3 @@
+from django.db import models
+
```
jfaracik: Empty

PR #32, opened 2019-06-18
Lt 919 define new work queues

anirbanlahiri-dit:
This is an early review. This Branch would need some more updates once LT-936 is merged to master.


queues/services.py
```diff
@@ -0,0 +1,22 @@
...
+    data = post(request, QUEUES_URL, json)
+    return data.json(), data.status_code
+
+
+def get_queue(request, pk):
+    data = get(request, QUEUES_URL + pk)
+    return data.json(), data.status_code
+
+
+def update_queue(request, pk, json):
```
jfaracik: Rename to _put_queue_ for consistency.

queues/views.py
```diff
@@ -0,0 +1,76 @@
...
+            'data': data,
+            'title': 'Queues',
+            'user_data': user_data,
+        }
+        return render(request, 'queues/index.html', context)
+
+
+class AddQueue(TemplateView):
+    def get(self, request, **kwargs):
+        context = {
```
jfaracik: Use the form_page function as opposed to rendering the form yourself:

eg return form_page(request, denial_reasons_form())

queues/views.py
```diff
@@ -0,0 +1,76 @@
...
+            'title': 'Add Queue',
+            'page': forms.form,
+        }
+        return render(request, 'form.html', context)
+
+    def post(self, request, **kwargs):
+        # including the user's team details in the post request
```
jfaracik: Use the submit_single_form function

An example can be seen in DenyCase

PR #35, opened 2019-06-19
fix: organisation detail

cstobart:


organisations/services.py
```diff
@@ -15,3 +15,8 @@ def post_organisations(request, json):
 def get_organisations_sites(request, pk):
     data = get(request, ORGANISATIONS_URL + pk + SITES_URL)
     return data.json(), data.status_code
+
+
+def get_organisation_detail(request, pk):
```
jfaracik: Rename to get_organisation

PR #56, opened 2019-07-08
feat: LT-945 - Attach documents to cases

jfaracik:
* Added page to view all documents on a case
* Added page to add multiple documents to a case
* Integrated with S3 to enable this


ui_automation_tests/step_defs/document_upload_test.py
```diff
@@ -0,0 +1,62 @@
...
+def click_attach_documents(driver):
+    documents_page = DocumentsPage(driver)
+    documents_page.click_attach_documents()
+
+@when('I upload a file')
+def upload_a_file(driver):
+    attach_document_page = AttachDocumentPage(driver)
+
+    # Path gymnastics to get the absolute path for $PWD/../resources/(file_to_upload_x) that works everywhere
+    file_to_upload_abs_path = \
```
phlptt: has this been run on the jenkins job yet? wondering if it will work there.

ui_automation_tests/pages/documents_page.py
```diff
@@ -0,0 +1,24 @@
+class DocumentsPage():
+
+    def __init__(self, driver):
+        self.driver = driver
+        self.attach_docs_btn = '.govuk-button[href*="attach"]' #css
+        self.latest_filename =  '//*[@id="main-content"]/div[2]/div[1]/p[1]' #xpath
+        self.second_latest_filename =  '//*[@id="main-content"]/div[3]/div[1]/p[1]' #xpath
+        self.latest_file_description =  '//*[@id="main-content"]/div[2]/div[1]/p[2]' #xpath
```
phlptt: is there another selector to get for these. as they are relative to the location on the page this may change

ui_automation_tests/features/document_upload.feature
```diff
@@ -0,0 +1,17 @@
+@internal
```
phlptt: can you add another tag here for something like @documents

ui_automation_tests/step_defs/document_upload_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+from ui_automation_tests.pages.application_page import ApplicationPage
+from ui_automation_tests.pages.attach_document_page import AttachDocumentPage
+from ui_automation_tests.pages.documents_page import DocumentsPage
+
+scenarios('../features/document_upload.feature', strict_gherkin=False)
+
+import logging
+log = logging.getLogger()
+console = logging.StreamHandler()
```
phlptt: doesnt look like this is used

ui_automation_tests/features/document_upload.feature
```diff
@@ -0,0 +1,17 @@
...
+
+@LT-945_upload
+Scenario: Upload a new document that doesn't contain a virus
+  Given I go to internal homepage
+  When I click on application previously created
+  And I click on the Documents button
+  And I click on the Attach Document button
+  And I upload a file
+  And I click on the Attach Document button
+  And I upload another file
```
phlptt: could you specify this here as upload file file-name-1 or something like that?

ui_automation_tests/features/document_upload.feature
```diff
@@ -0,0 +1,17 @@
...
+  Then the files are listed on the Documents page newest first
```
phlptt: can we do some negative tests here as well e.g. click on the submit button and get error message saying file not uploaded, and file over 100mb with error message too (not sure if this one will be possible)

ui_automation_tests/step_defs/document_upload_test.py
markm5A8Z9F43913A: Yes.

ui_automation_tests/features/document_upload.feature
markm5A8Z9F43913A: Done.

cases/middleware.py
```diff
@@ -0,0 +1,34 @@
+from django.shortcuts import redirect
```
cstobart: Why is this in the cases app?

templates/cases/case/application-case.html
```diff
@@ -2,7 +2,7 @@
 
 {% load humanize %}
 
-{% block back-link %}
+{% block back-button %}
```
cstobart: should be back link

ui_automation_tests/pages/documents_page.py
markm5A8Z9F43913A: Done

ui_automation_tests/step_defs/document_upload_test.py
markm5A8Z9F43913A: Removed

PR #57, opened 2019-07-08
Lt 1006 raise a clc

5pence:
please review


cases/views.py
```diff
@@ -70,6 +70,33 @@ def post(self, request, **kwargs):
         return redirect('/cases/' + case_id + '#case_notes')
 
 
+class ViewCLCCase(TemplateView):
+    def get(self, request, **kwargs):
+        case_id = str(kwargs['pk'])
+        case, status_code = get_case(request, case_id)
+        print(case)
```
jfaracik: print and commented out lines

andytaylorkainos: Nicely spotted, done

PR #65, opened 2019-07-16
fix: CLC Query redirects to the correct place now

jfaracik:


cases/views.py
```diff
@@ -80,7 +80,7 @@ def post(self, request, **kwargs):
             error = error.replace('this field', 'the case note')  # TODO: Move to API
             return error_page(request, error)
 
-        return redirect('/cases/' + case_id + '#case_notes')
+        return redirect(reverse('cases:case', kwargs={'pk': case_id}) + '#case_notes')
```
efahy: As suggested in Slack, there are a few other instances in the code where you might consider making a similar change for consistency

PR #67, opened 2019-07-18
Lt 912 add an external note to a case

5pence:


ui_automation_tests/conftest.py
```diff
@@ -17,18 +17,18 @@ def pytest_exception_interact(node, report):
...
 def pytest_addoption(parser):
     env = str(os.environ.get('ENVIRONMENT'))
     if env == 'None':
         env = "dev"
     print("touched: " + env)
     parser.addoption("--driver", action="store", default="chrome", help="Type in browser type")
-    parser.addoption("--exporter_url", action="store",
-                      default="https://exporter.lite.service." + env + ".uktrade.io/", help="url")
-    parser.addoption("--internal_url", action="store",
-    # parser.addoption("--exporter_url", action="store", default="http://localhost:8300", help="url")
```
jfaracik: Should this file be changed?

ui_automation_tests/conftest.py
```diff
@@ -17,18 +17,18 @@ def pytest_exception_interact(node, report):
         # utils.save_screenshot(node.funcargs.get("driver"), name)
 
 
-# Create driver and url command line addoption
+# Create driver and url command line adoption
 def pytest_addoption(parser):
     env = str(os.environ.get('ENVIRONMENT'))
     if env == 'None':
         env = "dev"
     print("touched: " + env)
```
cstobart: Remove or add to logging

ui_automation_tests/conftest.py
```diff
@@ -17,18 +17,18 @@ def pytest_exception_interact(node, report):
...
     parser.addoption("--driver", action="store", default="chrome", help="Type in browser type")
-    parser.addoption("--exporter_url", action="store",
-                      default="https://exporter.lite.service." + env + ".uktrade.io/", help="url")
-    parser.addoption("--internal_url", action="store",
-    # parser.addoption("--exporter_url", action="store", default="http://localhost:8300", help="url")
-    # parser.addoption("--internal_url", action="store", default="http://localhost:8080", help="url")
+    # parser.addoption("--exporter_url", action="store",
+    #                   default="https://exporter.lite.service." + env + ".uktrade.io/", help="url")
```
cstobart: reverse commenting so tests run against dev

ui_automation_tests/conftest.py
5pence: yes have changed it - good catch Jan

ui_automation_tests/conftest.py
5pence: done

ui_automation_tests/conftest.py
5pence: done good catch

efahy:
It looks like the guidance text around case note length is not appearing when entering the text of a new note - can you investigate please?

templates/cases/case/tabs/activity.html
```diff
@@ -84,19 +81,23 @@ <h3 class="govuk-heading-m">Goods Location</h3>
...
+					</div>
 					<a id="case-note-cancel-button" href="{% url 'cases:case' data.case.id %}" class="govuk-button govuk-button--secondary" type="button" draggable="false">
 						Cancel
 					</a>
-					<p id="case_note-warning" class="govuk-hint govuk-character-count__message" aria-live="polite">
-						You can enter up to 2200 characters
-					</p>
+<!-- {#					<p id="case_note-warning" class="govuk-hint govuk-character-count__message" aria-live="polite">#}
+{#						You can enter up to 2200 characters#}
+{#					</p>#} -->
```
efahy: Looks like this is why the note length text isn't appearing - its commented out?

5pence: This is now sorted out

5pence: ALSO character count is now working, and e2e test written

PR #72, opened 2019-07-22
refactor: docker config to support dev framework

mh-hmcts:


docker.env
```diff
@@ -1,21 +1,21 @@
...
 
 # AWS
-AWS_ACCESS_KEY_ID=''
-AWS_SECRET_ACCESS_KEY=''
-AWS_STORAGE_BUCKET_NAME=''
-AWS_REGION='eu-west-2'
+AWS_ACCESS_KEY_ID=
+AWS_SECRET_ACCESS_KEY=
+AWS_STORAGE_BUCKET_NAME=
+AWS_REGION=eu-west-2
```
markm5A8Z9F43913A: Do we want the value here specified?

PR #73, opened 2019-07-23
Setup hooks

phlptt:
Before hooks to set up organisations and applications when they are needed


README.md
```diff
@@ -54,17 +54,18 @@ Note the migrations will need to be run again the next time the service is to be
...
+* Setup chromedriver
+  * Download chromedriver from http://chromedriver.chromium.org/ and install it  
+  * make sure it has execute permissions and is in PATH
+* or
+  * `brew cask install chromedriver`
+
+* Setup dev pipenv environment:
   * `pipenv sync -d`
 * Run `pipenv run python -m pytest`
+* For a specific tag (dont use the @)  `py.test -k "tag name"`
```
andytaylorkainos: Can you check the difference between `-k` and `-m` (i think `-k` looks for filenames/regex where-as `-m` looks for annotations)

ui_automation_tests/conftest.py
```diff
@@ -23,20 +28,20 @@ def pytest_exception_interact(node, report):
...
+        env = "staging"
     print("touched: " + env)
     parser.addoption("--driver", action="store", default="chrome", help="Type in browser type")
     parser.addoption("--exporter_url", action="store",
                      default="https://exporter.lite.service." + env + ".uktrade.io/", help="url")
     parser.addoption("--internal_url", action="store",
                      default="https://internal.lite.service." + env + ".uktrade.io/", help="url")
-    # parser.addoption("--exporter_url", action="store", default="http://localhost:8300", help="url")
-    # parser.addoption("--internal_url", action="store", default="http://localhost:8200", help="url")
+    #parser.addoption("--exporter_url", action="store", default="http://localhost:8300", help="url")
```
andytaylorkainos: This is terribly OCD of me, but the space after the `#` is standard for PEP8 - can you add it back please :)

ui_automation_tests/conftest.py
```diff
@@ -93,6 +98,143 @@ def invalid_username():
 sso_password = env('TEST_SSO_PASSWORD')
 
 
+@pytest.fixture(scope="session")
```
andytaylorkainos: I like the fixture plan, but can we remove the code to its own file? and import it at the top of `conftest.py` so that we don't end up with one file over 1k lines long

ui_automation_tests/helpers/helpers.py
```diff
@@ -108,3 +108,29 @@ def apply_style(s):
     apply_style("background: yellow; border: 2px solid red;")
     time.sleep(.7)
     apply_style(original_style)
+
+
+def get_element_by_text(elements, text: str):
```
andytaylorkainos: 👍

ui_automation_tests/helpers/helpers.py
```diff
@@ -108,3 +108,29 @@ def apply_style(s):
...
+    """
+    Loops through the list of elements, checks if the text is equal to
+    text and returns the element if so
+    """
+    for element in elements:
+        if element == text:
+            return element
+
+
+def get_element_index_by_text(elements, text: str):
```
andytaylorkainos: 👍

ui_automation_tests/helpers/helpers.py
efahy: These two new helper methods don't seem to be used.  Do we need them? I worry slightly about the concept of looping through elements to find a match - Is that not doable using xpaths or some other standard method?

ui_automation_tests/step_defs/organisation_test.py
```diff
@@ -0,0 +1,112 @@
...
+    def verify_registered_organisation(driver):
+        if not context.org_registered_status:
+            exists = utils.is_element_present(driver, By.XPATH, "//*[text()[contains(.,'" + context.org_name + "')]]")
+            assert exists
+            registration_complete_message = driver.find_element_by_tag_name("h1").text
+            assert registration_complete_message == "Organisation Registered"
+
+
+    @when('I go to organisations')
+    def i_go_to_organisations(driver):
```
efahy: I've seen in some places that we ditch `i_` off the start of the step method, even though its in the feature step name.  If that is a convention we're using it needs to be applied here

ui_automation_tests/step_defs/organisation_test.py
```diff
@@ -0,0 +1,112 @@
...
+                organisations_form_page.enter_name(name)
+                context.org_name = name
+            else:
+                context.org_name = name+utils.get_formatted_date_time_m_d_h_s()
+                organisations_form_page.enter_name(context.org_name)
+            organisations_form_page.enter_eori_number(eori)
+            organisations_form_page.enter_sic_number(sic)
+            organisations_form_page.enter_vat_number(vat)
+            organisations_form_page.enter_registration_number(registration)
+            organisations_form_page.click_submit()
```
efahy: Some methods here use `OrganisationsPage` and others don't - they probably all should?

ui_automation_tests/helpers/helpers.py
phlptt: it's being used in future branches, we have been previously been using xpaths but that's what made everything break recently.

ui_automation_tests/step_defs/organisation_test.py
phlptt: wouldnt say theres a convention set for this. but happy to have it as one!

README.md
```diff
@@ -54,17 +54,18 @@ Note the migrations will need to be run again the next time the service is to be
...
+* You may need to make sure in pycharm, within Preferences -> Tools -> Python Integrated Tools -> Default Test Runner is pytest
```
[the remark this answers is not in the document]
phlptt: i'll write up a bit more info on confluence at some point and then link to it if thats alright?

ui_automation_tests/conftest.py
```diff
@@ -131,16 +278,25 @@ def go_to_exporter_when(driver, exporter_url):
...
+def login_to_exporter(driver, exporter_url, username, password, set_up_org):
+    driver.get(exporter_url)
     if username == "TestBusinessForSites@mail.com":
         username = context.email
     exporter_hub = ExporterHub(driver)
     if "login" in driver.current_url:
         exporter_hub.login(username, password)
 
 
+@when(parsers.parse('I login to exporter homepage with username context username and "{password}"'))
```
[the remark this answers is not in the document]
phlptt: yea i did it delibaretely to follow on from a specific test

ui_automation_tests/step_defs/test_00_login.py
```diff
@@ -30,5 +30,6 @@ def login_button_click(driver):
 
 @then('I see you need to sign in error message')
 def login_error_message(driver):
+    log.info("asdsadsadas")
```
andytaylorkainos: Remove this line, also - did it log anything? :D

PR #76, opened 2019-07-29
Move cases

phlptt:


ui_automation_tests/helpers/helpers.py
```diff
@@ -137,8 +137,11 @@ def get_element_index_by_text(elements, text: str):
 
 
 def wait_until_page_is_loaded(driver):
-    while True:
+    time_no = 0
+    while time_no < 60:
         if driver.execute_script("return document.readyState") == "complete":
+            time.sleep(1)
```
andytaylorkainos: Need to move these `time` operations out of the `if` condition so that the timer keeps going up even when the `document.readyState` is not complete :)

PR #77, opened 2019-07-31
feat: LT-1077 - Picklist

jfaracik:
* Added pick list to current team
* Add items
* Edit items


teams/urls.py
```diff
@@ -4,12 +4,15 @@
...
-	path('add', views.AddTeam.as_view(), name='add'),
-	# ex: /teams/43a88949-5db9-4334-b0cc-044e91827451/
-	path('<uuid:pk>', views.TeamDetail.as_view(), name='team'),
-	# ex: /teams/43a88949-5db9-4334-b0cc-044e91827451/edit
-	path('<uuid:pk>/edit', views.EditTeam.as_view(), name='edit'),
+    # ex: /team/
+    path('', views.Team.as_view(), name='team'),
+
+    # ex: /teams/
+    path('s/', views.TeamsList.as_view(), name='teams'),
```
andytaylorkainos: howcome we need a leading `s/` in the url?

jfaracik: As the base url is team, we had an`s/` to make it teams - that way we can have team and teams in the same urls.py

andytaylorkainos: Do we need a `/team/` url - why cant we use `/teams/` as the base and use `/teams/<uuid:pk>/ for the `team` endpoint 🤔

ui_automation_tests/pages/picklist_pages.py
```diff
@@ -0,0 +1,48 @@
+class PicklistPages():
+
+    def __init__(self, driver):
+        self.driver = driver
+        self.picklist_tab = '.lite-tabs__tab[href*="picklists"]' #css
+        self.picklist_edit_button = '.govuk-button[href*="edit"]' #css
```
andytaylorkainos: this is picky, but can you put two spaces between the comments and the code, and also separate the comment text by one space from the tag:

E.G
`self.picklist_tab = '.lite-tabs__tab[href*="picklists"]'  # css`

PR #79, opened 2019-08-01
feat: LT-1123 View all cases

markm5A8Z9F43913A:


ui_automation_tests/features/queues.feature
```diff
@@ -79,4 +79,48 @@ Feature: I want to define new work queues and the teams they belong to
...
+    And I click on add a queue
+    And I enter in queue name "Additional queue"
+    And I go to the internal homepage
+    Then I see previously created application
+    When I click on new queue in dropdown
+    Then There are no cases shown
+#    When I click on All cases queue in dropdown
+#    Then I see previously created application
+#    When I click on Open cases queue in dropdown
+#    Then I see previously created application
```
markm5A8Z9F43913A: We're aware of these commented out lines, please ignore them for the moment. Will be removed before merging.

ui_automation_tests/conftest.py
```diff
@@ -161,3 +166,44 @@ def click_on_clc_case_previously_created(driver, context):
     case_list_page = CaseListPage(driver)
     assert case_list_page.assert_case_is_present(context.case_id)
     driver.find_element_by_css_selector('.lite-cases-table').find_element_by_xpath("//*[text()[contains(.,'" + context.case_id + "')]]").click()
+
+
+@when('I click record decision')
```
andytaylorkainos: are these step-defs needed in `conftest.py` - can they not be moved to their corresponding test step-defs?

markm5A8Z9F43913A: They are, because they need to be shared between two different test_*.py files. And conftest.py is the preferred shared location from my understanding. If you have a suggestion where else to put it, I'd be happy to do that.

andytaylorkainos: Was only making sure :) if theyre shared then i think thats the best place to put them for now.

Nick and I were thinking of how we could create a shared step-def directory but it doesnt seem to import particularly well for whatever reason :(

PR #82, opened 2019-08-02
refactor: speed up test framework

mh-hmcts:


ui_automation_tests/fixtures/core.py
```diff
@@ -12,15 +20,20 @@ def driver(request):
     chrome_options = webdriver.ChromeOptions()
     # remove this line to see it running in browser.
     # chrome_options.add_argument('--headless')
+    if str(env('TEST_TYPE')) == 'headless':
```
phlptt: this could do with getting added into local.env and docker.env as will error if there's nothing in peoples .env files

ui_automation_tests/helpers/seed_data.py
```diff
@@ -0,0 +1,180 @@
+import json
+import requests
+from conf.settings import env
+
+
+class SeedData:
+    base_url = 'http://localhost:8100'
```
phlptt: this will have to be changed to come from an environment variable as it wont work on other environments or if someones running on a different port

ui_automation_tests/helpers/seed_data.py
```diff
@@ -0,0 +1,180 @@
...
+    gov_user_email = 'test-uat-user@digital.trade.gov.uk'
```
phlptt: get this from env file

ui_automation_tests/helpers/seed_data.py
```diff
@@ -0,0 +1,180 @@
...
+        "first_name": "ecju",
+        "last_name": "user"}
+    gov_headers = {'content-type': 'application/json'}
+    export_headers = {'content-type': 'application/json'}
+    context = {}
+    logging = True
+
+    request_data = {
+        "organisation": {
+            "name": "ExporterOrg",
```
phlptt: Can you change this to use context.org_name as it is hardcoded in some places as 'Test Org' 😱

mh-hmcts: OK

ui_automation_tests/fixtures/core.py
mh-hmcts: agreed

ui_automation_tests/helpers/seed_data.py
mh-hmcts: done

PR #89, opened 2019-08-08
feat: LT-1376 - User Advice

jfaracik:
* Added {% include %} views for displaying case data, with optional `show_checkboxes` and `show_advice` fields
* Updated to work with the latest API changes
* Added screens for giving advice for goods/destinations
* Added `pretty_json`, use it to pretty print JSON - super helpful for developers, **don't** use it for non development purposes


cases/services.py
```diff
@@ -1,6 +1,6 @@
 from conf.client import post, get, put, delete
 from conf.constants import CASE_URL, CASE_NOTES_URL, APPLICATIONS_URL, ACTIVITY_URL, CLC_QUERIES_URL, DOCUMENTS_URL, \
-    CASE_FLAGS_URL
+    CASE_FLAGS_URL, ADVICE_URL
 
 
 def get_case(request, pk):
```
5pence: This case and the overloaded one underneath is not needed in Python. Use *args and write once.

cases/services.py
```diff
@@ -9,19 +9,19 @@ def get_case(request, pk):
 
 
 def put_case(request, pk, json):
-    data = put(request, CASE_URL + pk + '/', json)
+    data = put(request, CASE_URL + pk, json)
```
5pence: see comment above

cases/views/advice.py
```diff
@@ -0,0 +1,161 @@
...
+                'goods_types': data.get('goods_types'),
+                'countries': data.get('countries'),
+                'end_user': data.get('end_user'),
+                'ultimate_end_users': data.get('ultimate_end_users'),
+                'errors': response['errors'],
+                'data': data,
+            }
+            return render(request, self.form, context)
+
+        return success_page(request,
```
andytaylorkainos: Can you specify which parameter each of these arguments is for? `title=` etc

templates/cases/case/clc-query-case.html
```diff
@@ -25,20 +25,8 @@ <h1 class="govuk-heading-l">{{ title }}</h1>
...
-            {{ flag.name }}
-        </div>
-    {% endfor %}
-</div>
-
-<div id="flags-modal" style="display: none;">
-	{% include "cases/case/flags.html" %}
-</div>
+<!-- Case flags -->
+{% include 'cases/case/views/case-flags.html' %}
```
andytaylorkainos: Nice!

PR #126, opened 2019-09-16
Linting

lukesa:


ui_automation_tests/conftest.py
```diff
@@ -35,7 +32,7 @@ def pytest_addoption(parser):
     if env == 'local':
         parser.addoption("--internal_url", action="store", default="http://localhost:8200", help="url")
         parser.addoption("--lite_api_url", action="store", default="http://localhost:8100", help="url")
-    elif env == 'dev2':
+    elif env == 'dev2' or env == 'DEV2':
```
andytaylorkainos: can we just do `env.lower() == `dev2`?

PR #132, opened 2019-09-23
feat: LT-1486 - Add Additional Details for End User Advisory

cstobart:


templates/cases/case/queries/end_user_advisory.html
```diff
@@ -86,19 +148,6 @@ <h3 class="govuk-heading-m">End User Details</h3>
 			</div>
 		</dl>
 
-		{% if case.query.copy_of %}
```
jfaracik: Why was this removed?

PR #136, opened 2019-09-24
feat: LT-1417 - Individual Exporter Account

cstobart:


register_business/forms.py
```diff
@@ -1,51 +1,92 @@
...
 
-def register_business_forms():
+def conditional(condition: bool, obj, obj_2=None):
+    """
+    Returns the object depending on a condition
+    """
+    if condition:
+        return obj
+    elif obj_2:
+        return obj_2
```
BenAndrew-Kainos: Is it ok that if condition is False and obj_2 is None that nothing is returned. (Probably fine just making sure)

register_business/forms.py
```diff
@@ -1,51 +1,92 @@
...
+                                       name='user.email'),
+                             TextInput(title='First name',
+                                       name='user.first_name'),
+                             TextInput(title='Last name',
+                                       name='user.last_name'),
+                             TextInput(title='European Union registration and identification number (EORI)',
+                                       optional=True,
+                                       name='eori_number'),
+                             TextInput(title='UK VAT number',
+                                       description='9 digits long, with the first two letters indicating the'
```
BenAndrew-Kainos: Looks like a lot of strings that could be moved to strings.json... :D

register_business/forms.py
```diff
@@ -1,51 +1,92 @@
...
+                                             TextInput(title='Last name',
+                                                       name='user.last_name'),
+                                         ],
+                                         default_button_name='Submit',
+                                         helpers=[
+                                             HelpSection('Help', 'This will be the default user for this organisation.')
+                                         ]))
     ],
-        show_progress_indicators=True,)
+        show_progress_indicators=True, )
```
andytaylorkainos: please remove white space after  `show_progress_indicators=True`

register_business/forms.py
cstobart: Yes deliberate that there is no return otherwise

register_business/forms.py
cstobart: (nobody has used those in months...)

register_business/forms.py
cstobart: of course

register_business/forms.py
BenAndrew-Kainos: Think you find we have, and half my code review comments on my PR's are for lite-content, but you seem to dodge it

BenAndrew-Kainos: think it's time to bite the bullet and start using it. hat or we all agree it either is or isn't being used

stephanieboyd: I agree with Ben, these should be moved to lite-content. If as a team we want to move away from the strings.json approach, we should discuss and move forward but until then let's keep it consistent :)

cstobart: London has the opposite opinion, but we'll ask Nicolette tomorrow which she'd prefer, I'll do these ones for now

cstobart: done

ui_automation_tests/features/organisation.feature
```diff
@@ -9,7 +9,18 @@ Feature: I want to add a company to LITE
     Given I go to internal homepage
     When I go to organisations
     And I choose to add a new organisation
+    And I select "commercial"
     And I provide company registration details of name: "BlueOcean", EORI: "GB987654312000", SIC: "73200", VAT: "123456789", CRN: "000000011"
     And I setup an initial site with name: "HQ", addres line 1: "123 Cobalt Street", town or city: "London", County: "Islington", post code: "AB1 2CD", country: "Ukraine"
     And I setup the admin user with email: "TestBusinessForSites@mail.com", first name: "Trinity", last name: "Fishburne"
     Then organisation is registered
+
+  @LT_1417_test
```
phlptt: can you name this to something a bit more meaningful

ui_automation_tests/features/organisation.feature
```diff
@@ -9,7 +9,18 @@ Feature: I want to add a company to LITE
...
+  Scenario: Test registering and individual
```
phlptt: typo in 'and' should be an

PR #146, opened 2019-09-27
Added support for a proxy while running e2e tests

Lightman32:
Adding a proxy support for E2E tests. This will enable us to use ZAP proxy and get some security coverage.


ui_automation_tests/fixtures/core.py
```diff
@@ -28,6 +28,10 @@ def driver(request):
         chrome_options.add_argument('--headless')
     chrome_options.add_argument('--no-sandbox')
 
+# Use proxy settings provided in config file for security testing
```
andytaylorkainos: can you indent the comment (i know its super minor)

Lightman32: done, thanks

PR #149, opened 2019-09-30
test: LT-958 - end to end test for moving case and queue change appearing in audit trail

phlptt:
Feature already merged to master
Test checks the text "moved the case to: queuename" is the first item in the audit trail straight after a case move.

### DIFF
--- a/ui_automation_tests/step_defs/test_queues.py
+++ b/ui_automation_tests/step_defs/test_queues.py
@@ -77,9 +77,6 @@ def deselect_all_queues(driver):
     Shared(driver).click_submit()
 
 
-@when('I move case to new cases original queue and remove from new queue')
-def move_case_to_original_queue(driver, context):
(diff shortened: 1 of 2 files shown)


ui_automation_tests/step_defs/test_queues.py
```diff
@@ -77,9 +77,6 @@ def deselect_all_queues(driver):
     Shared(driver).click_submit()
 
 
-@when('I move case to new cases original queue and remove from new queue')
-def move_case_to_original_queue(driver, context):
-    ApplicationPage(driver).click_move_case_button()
-    driver.find_element_by_id(context.queue_name).click()
-    QueuesPages(driver).click_on_new_cases_queue()
-    Shared(driver).click_submit()
+@then("queue change is in audit trail")
```
stephanieboyd: Maybe you want to make this more specific, "queue change is first in the audit trail"?

PR #150, opened 2019-09-30
Lt 1334 goods country decision matrix

cstobart:


templates/cases/case/finalise-open-goods-countries.html
```diff
@@ -0,0 +1,84 @@
...
+
+	{% if errors %}
+		<div class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary">
+			<h2 class="govuk-error-summary__title" id="error-summary-title">
+				There is a problem
+			</h2>
+			<div class="govuk-error-summary__body">
+				<ul class="govuk-list govuk-error-summary__list">
+					<li>
+						<a href="#">Select a decision for each good and country combination</a>
```
jfaracik: Link to first radio button

cases/views/advice.py
```diff
@@ -199,20 +201,137 @@ def post(self, request, **kwargs):
...
+    def post(self, request, *args, **kwargs):
+        case, data, keys = self._generate_data(request, str(kwargs['pk']))
+
+        data = request.POST.copy()
+        data.pop('csrfmiddlewaretoken')
+        selection = {}
+        action = data.pop('action')[0]
+
+        selection['good_countries'] = []
+        for key, _ in data.items():
```
BenAndrew-Kainos: if you don't need values from the dict just use data.keys()

cases/views/advice.py
```diff
@@ -199,20 +201,137 @@ def post(self, request, **kwargs):
...
+                {
+                    'case': str(kwargs['pk']),
+                    'good': key.split('.')[0],
+                    'country': key.split('.')[1],
+                    'decision': data.get(key)}
+            )
+
+        context = {
+            'case': case,
+            'decisions': ['approve', 'refuse', 'no_licence_required'],
```
BenAndrew-Kainos: same as above comment

cases/views/advice.py
```diff
@@ -199,20 +201,137 @@ def post(self, request, **kwargs):
...
+
+        # If errors, return page
+        if context['errors']:
+            context['good_countries'] = post_data
+            return render(request, 'cases/case/finalise-open-goods-countries.html', context)
+
+        data, _ = post_good_countries_decisions(request, str(kwargs['pk']), selection)
+
+        if action == 'save':
+            print(action)
```
BenAndrew-Kainos: remove print

cases/views/advice.py
```diff
@@ -199,20 +201,143 @@ def post(self, request, **kwargs):
...
+        # The keys are each relevant good-country pairing in the format good_id.country_id
+        keys = []
+        # Builds form page data structure
+        # For each good in the case
+        for good in case['application']['goods_types']:
+            # Match the goods with the goods in advice for that case
+            # and attach the advice value to the good
+            for advice in _advice['advice']:
+                if advice['goods_type'] == good['id']:
+                    good['advice'] = advice['type']
```
a-bek: Break out of the loop?

cases/views/advice.py
```diff
@@ -199,20 +201,143 @@ def post(self, request, **kwargs):
...
+            # If the good has countries attached to it as destinations
+            # We do the same with the countries and their advice
+            if good['countries']:
+                for country in good['countries']:
+                    keys.append(str(good['id']) + '.' + country['id'])
+                    for advice in _advice['advice']:
+                        if advice['country'] == country['id']:
+                            country['advice'] = advice['type']
```
a-bek: Break out of loop?

cases/views/advice.py
```diff
@@ -199,20 +201,143 @@ def post(self, request, **kwargs):
...
+            else:
+                good['countries'] = []
+                # We attach all countries from the case
+                # And then attach the advice as before
+                for country in case['application']['destinations']['data']:
+                    good['countries'].append(country)
+                    keys.append(str(good['id']) + '.' + country['id'])
+                    for advice in _advice['advice']:
+                        if advice['country'] == country['id']:
+                            country['advice'] = advice['type']
```
a-bek: Break out of loop?

cases/views/advice.py
```diff
@@ -199,20 +201,143 @@ def post(self, request, **kwargs):
...
+        data = get_good_countries_decisions(request, pk)
+        print('data1', data)
```
a-bek: delete the print

cases/views/advice.py
```diff
@@ -199,20 +201,143 @@ def post(self, request, **kwargs):
...
+        if 'detail' in data:
+            return error_page(request, 'You do not have permission.')
+
+        return case, data, keys
+
+    @staticmethod
+    def _generate_errors(keys, request_data, action):
```
a-bek: Rename `_generate_post_data_and_errors'?

cases/views/advice.py
```diff
@@ -199,20 +201,143 @@ def post(self, request, **kwargs):
         return post_advice_details(post_final_case_advice, request, self.case, self.form, 'final')
 
 
+class FinaliseGoodsCountries(TemplateView):
+    @staticmethod
+    def _generate_data(request, pk):
```
a-bek: This is fine for now but generally this kind of logic should be put into like a `service.py` file (or something similar) within the app that handles the business logic of the views, broken down into logical components so it's easier to read and test! it should also abstract away django specifics (passing the request object).

cases/views/advice.py
a-bek: value is needed below: `'decision': data.get(key)` -> `'decision': value`

cases/views/advice.py
```diff
@@ -199,20 +201,137 @@ def post(self, request, **kwargs):
...
+                                  'country': country_pk,
+                                  'decision': data.get(key)})
+        return context, post_data
+
+    def get(self, request, *args, **kwargs):
+        case, data, _ = self._generate_data(request, str(kwargs['pk']))
+        context = {
+            'case': case,
+            'good_countries': data['data'],
+            'decisions': ['approve', 'refuse', 'no_licence_required'],
```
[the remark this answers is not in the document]
cstobart: They're not contained in the model, and there are only two more to add in the future (they are the equivalent of yes/no/null)

cases/views/advice.py
```diff
@@ -199,20 +201,137 @@ def post(self, request, **kwargs):
...
+                        if advice['country'] == country['id']:
+                            country['advice'] = advice['type']
+            # If the good has no countries:
+            else:
+                good['countries'] = []
+                # We attach all countries from the case
+                # And then attach the advice as before
+                for country in self.case['application']['destinations']['data']:
+                    good['countries'].append(country)
+                    self.keys.append(str(good['id']) + '.' + country['id'])
```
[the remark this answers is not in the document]
cstobart: not viable

cases/views/advice.py
cstobart: 👍

ui_automation_tests/features/give_advice.feature
```diff
@@ -94,3 +94,29 @@ Feature: I want to record my user advice and any comments and conditions relatin
...
+    And I go to the team advice
+    And I combine all advice
+    And I go to the final advice
+    And I combine all advice
+    And I finalise the goods and countries
+    And I click continue
+    Then I see error message "Albania (Approve)"
+    When I select approve for all combinations of goods and countries
+    And I click continue
+    Then I reset the permissions
```
phlptt: might be worth having a step to verify that this approval has worked, guess all that can be done is check the status is finalised and status has changed though

cases/views/advice.py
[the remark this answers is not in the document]
BenAndrew-Kainos: I would disagree. Perhaps see what over dev's think, can they follow it easily without any prior explanation?

cases/views/advice.py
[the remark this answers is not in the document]
BenAndrew-Kainos: Ok could you define it as a constant then to be shared across applicable views

cases/views/advice.py
BenAndrew-Kainos: oh ok so change _ to value and change data.get(key) to value

ui_automation_tests/features/give_advice.feature
cstobart: That's a pretty transient check as it is, so waiting until there is actually something verifiable to test it against aka licences etc.

cstobart: These checks will be done in the upcoming stories where we lock down cases once they're 'done'

cases/views/advice.py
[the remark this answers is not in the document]
cstobart: sure, its unique to this view, but I will do

cases/views/advice.py
cstobart: that was already done ^outdated

cases/views/advice.py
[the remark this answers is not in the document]
cstobart: 👍

cases/views/advice.py
BenAndrew-Kainos: wth I can't see it :(

BenAndrew-Kainos:
Github is being very odd and I can't see your changes. Approving as you've verified these changes have been made

PR #156, opened 2019-10-02
feat: LT-1008 - HMRC Organisation

cstobart:


organisations/urls.py
```diff
@@ -9,4 +9,6 @@
     path('', views.OrganisationList.as_view(), name='organisations'),
     # ex: /43a88949-5db9-4334-b0cc-044e91827451/
     path('<uuid:pk>/', views.OrganisationDetail.as_view(), name='organisation'),
+    # ex: /hmrc/
```
markm5A8Z9F43913A: Wrong example

ui_automation_tests/pages/header_page.py
```diff
@@ -15,6 +15,9 @@ def click_lite_menu(self):
     def click_organisations(self):
         self.driver.find_element_by_css_selector("a[href*='/organisations/']").click()
 
+    def click_hmrc(self):
+        self.driver.find_element_by_css_selector("a[href*='/hmrc/']").click()
```
markm5A8Z9F43913A: I'm not bothered by this personally if it's used in one place, but I was yelled at in the past for not storing the selectors in their own variables.

ui_automation_tests/step_defs/test_organisation.py
```diff
@@ -92,3 +100,20 @@ def fill_out_admin_user_details(driver, email, first_name, last_name, context):
         organisations_form_page.enter_first_name(first_name)
         organisations_form_page.enter_last_name(last_name)
         organisations_form_page.click_submit()
+
+
+@when(
+    parsers.parse(
+        'I provide hmrc registration details of org_name: "{org_name}", site_name: "{site_name}", addres line 1: "{address}", town or city: "{city}", County: "{region}", post code: "{post_code}", country: "{country}"'))
```
markm5A8Z9F43913A: Is it possible to break this line into two? It seems a bit long and I don't like having to scroll right in order to read it

ui_automation_tests/pages/header_page.py
cstobart: following the pattern of the file here, but normally I would put them in the init

register_business/forms.py
```diff
@@ -86,6 +86,36 @@ def register_business_forms(individual=False):
...
-                                         ]))
-    ],
-    show_progress_indicators=True)
+                                         ]))],
+        show_progress_indicators=True)
+
+
+def register_hmrc_organisation_forms(name='this organisation'):
+    return FormGroup([
+        Form(title='Register a HMRC organisation',
```
markm5A8Z9F43913A: an

ui_automation_tests/step_defs/test_organisation.py
cstobart: 👍

PR #159, opened 2019-10-07
LT-1584-bug-access-goods

BenAndrew-Kainos:


templates/cases/case/views/goods-flags.html
```diff
@@ -5,7 +5,7 @@
 		</div>
 	{% endfor %}
 	{% if not hide_creation_link %}
-		<a class="app-flag app-flag--action" href="{% url 'cases:assign_flags' case.id %}?level=goods&items={{ good.id }}&origin={{ origin }}">
+		<a class="app-flag app-flag--action" href="{% url 'cases:assign_flags' case_id %}?level=goods&items={{ good.id }}&origin={{ origin }}">
```
jfaracik: Why are we moving back to the case_id variable?

BenAndrew-Kainos: No idea. just adjusting for what the view actually passes in context

PR #171, opened 2019-10-09
feat: LT-1561 - CSP security

a-bek:
The Content Security Policy (CSP) defined allows us to control the resources we allow to be loaded into our frontend by whitelisting any urls and passing the information to the browser via the response header.


Pipfile
```diff
@@ -29,6 +29,7 @@ s3chunkuploader = "*"
 boto3 = ">=1.9"
 prospector = "*"
 lite-forms = "*"
+django-csp = "==3.5"
```
craiga: Is there a reason we're locking to a specific version here?

conf/settings.py
```diff
@@ -231,3 +232,12 @@
...
 SECURE_BROWSER_XSS_FILTER = True
+
+
+# CSP Security settings
+
+CSP_DEFAULT_SRC = ("'none'",)
+CSP_STYLE_SRC = ("'self'",)
+CSP_SCRIPT_SRC = ("'self'", 'ajax.googleapis.com')
+CSP_FONT_SRC = ("'self'",)
+CSP_IMG_SRC = ("'self'",)
```
craiga: I've seen lots of inline scripts and styles interleaved throughout the site. AFAIK this will break those.

Pipfile
a-bek: This is the latest version on PyPi. I prefer explicitly defining a version, and upgrading it when necessary.

conf/settings.py
a-bek: I'm running through now to try to collect all of them

conf/settings.py
```diff
@@ -228,8 +229,19 @@
...
 # Enable security features in hosted environments.
 SECURE_BROWSER_XSS_FILTER = not DEBUG
 SESSION_COOKIE_SECURE = not DEBUG
 CSRF_COOKIE_SECURE = not DEBUG
 SECURE_CONTENT_TYPE_NOSNIFF = not DEBUG
+
+# Content Security Policy
+CSP_DEFAULT_SRC = ("'self'",)
+CSP_STYLE_SRC = ("'self'", "'unsafe-inline'")
+CSP_SCRIPT_SRC = ("'self'", 'https://www.googletagmanager.com', 'https://script.hotjar.com', 'ajax.googleapis.com', 'https://www.google-analytics.com')
```
craiga: Does `ajax.googleapis.com` work as it's listed without the scheme prefix?

a-bek: Yes it does

PR #176, opened 2019-10-11
feat: LT-1505 - Change advice field labelling

craiga:
Changes advice field labelling.

This involves changes to each of the forms linked to from this page.

What do you advise

# Grant the licence

Grant the licence

# Add a proviso

Add a proviso

# Refuse the licence

Refuse the licence

# Tell the applicant they do not need a licence

Tell the applicant they do not need a licence

# Not applicable

Not applicable

This also changes some of the labels on the user advice page, where I've also fixed a styling issue which caused multiple lines of text to be displayed on top each other.

User advice page


templates/cases/case/views/other-documents.html
```diff
@@ -19,7 +19,7 @@ <h3 class="govuk-heading-m">Other Documents</h3>
                 </td>
                 <td class="lite-table__cell">
                     {% if additional_document.safe == True %}
-                    <a class="govuk-link lite-link--icon govuk-link--no-visited-state" id="additional_document" href="{% url 'cases:document' case.id additional_document.id %}">
```
craiga: Removing `lite-link--icon` class as it was causing multiple lines of text to wrap over themselves.

User advice page (with styling issue)

templates/cases/case/views/table-advice.html
```diff
@@ -29,17 +29,17 @@
                     <p class="govuk-body-s">{{ advice.proviso|default:'N/A' }}</p>
                 {% endif %}
                 {% if advice.denial_reasons|length %}
-                <p class="govuk-heading-s">Denial Reasons</p>
+                <p class="govuk-heading-s">Denial reasons</p>
```
craiga: Lowercased to make consistent with suggested labels from story.

templates/cases/case/give-advice.html
```diff
@@ -136,17 +145,20 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
...
 				</span>
 			{% endif %}
-<!--			<span class="govuk-hint" for="{{ question.name }}">-->
-<!--				This is a really long description-->
-<!--			</span>-->
+      <span class="govuk-hint" for="{{ question.name }}">
+        Optional
+      </span>
+      <span class="govuk-hint" for="{{ question.name }}">
+        This will appear on the licence
```
markm5A8Z9F43913A: We used to have the convention of storing the text strings in a shared submodule. Not sure what happened to the convention, but might be worth checking with @jfaracik because you have the same string appearing twice.

craiga: No problem, but what's the rationale behind this? Would this be better served by using [Django's built-in translation functions](https://docs.djangoproject.com/en/stable/topics/i18n/translation/)?

templates/cases/case/views/other-documents.html
andytaylorkainos: 👍

templates/cases/case/views/table-advice.html
andytaylorkainos: 👍

PR #177, opened 2019-10-11
Refactor shared fixtures

BenAndrew-Kainos:


BenAndrew-Kainos:
> Same comment from the other PR regarding `#noqa` applies here too.

Yeah. Have a look at the comment I left in that PR

PR #184, opened 2019-10-17
feat: LT-1580 - Amend Goods Permission

Mark-Spiers-kainos:


cases/views/goods.py
```diff
@@ -26,6 +27,10 @@ class ReviewGoods(TemplateView):
     def get(self, request, **kwargs):
         case_id = str(kwargs['pk'])
 
+        permissions = get_user_permissions(request, True)[0]
```
cstobart: does get_user_permissions return a list?

cases/views/clc_query.py
```diff
@@ -17,9 +18,14 @@ class Respond(TemplateView):
     form = None
 
     def dispatch(self, request, *args, **kwargs):
-        self.case = get_case(request, str(kwargs['pk']))
+        case_id = str(kwargs['pk'])
+        self.case = get_case(request, case_id)
         self.form = respond_to_clc_query_form(request, self.case)
 
+        permissions = get_user_permissions(request)
+        if 'ASSESS_GOODS' not in permissions:
```
stephanieboyd: As we discussed, let's not hardcode this 'ASSESS_GOODS' string here. Something shareable between api/frontends in lite content would be best. This will save us having to change this string in lots of places if something changes in the future (and potentially missing some).

ui_automation_tests/conftest.py
```diff
@@ -215,3 +217,23 @@ def assign_flags_to_case(driver, context):
...
+def get_required_permissions(driver):
+    roles_page = RolesPages(driver)
+    HeaderPage(driver).open_users()
+    UsersPage(driver).click_on_manage_roles()
+    roles_page.click_edit_for_default_role()
+    roles_page.edit_default_role_to_have_all_permissions()
+    Shared(driver).click_submit()
+
+
+@then("I reset the permissions") # noqa
```
phlptt: should be a when

cases/views/clc_query.py
Mark-Spiers-kainos: I will look at implementing it in lite_content when the move over from strings.json to a constants

PR #186, opened 2019-10-17
Ensure letter templates are valid Django templates.

craiga:
Screenshot 2019-10-17 at 16 13 35

### DIFF
--- a/picklists/views.py
+++ b/picklists/views.py
@@ -1,7 +1,9 @@
+from django import template
 from django.http import Http404
 from django.shortcuts import render, redirect
 from django.urls import reverse_lazy
 from django.views.generic import TemplateView
+
 from lite_forms.generators import form_page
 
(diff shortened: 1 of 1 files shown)


picklists/views.py
```diff
@@ -86,6 +88,23 @@ def get(self, request, **kwargs):
         return form_page(request, self.form, data=self.picklist_item)
 
     def post(self, request, **kwargs):
+        # Letter paragraphs are passed through the Django template engine, so we need
+        # to make sure they're valid.
+        if request.POST.get("type") == "letter_paragraph":
+            template_engine = template.engine.Engine.get_default()
+            try:
+                template_engine.from_string(request.POST["text"])
```
a-bek: Potential keyerror if submitted with empty `text`?

Otherwise lg2m :)

craiga: @a-bek Figure it's better to raise a `KeyError` than to silently accept a post with missing text.

PR #187, opened 2019-10-21
refactor: Update interface to match internal GOV requirements

jfaracik:
* Use Department for International Trade brand colour instead of the traditional GOV blue
* Use Arial/Sans-Serif instead of GDS Transport 
* Removed use of crown

image


templates/layouts/base.html
```diff
@@ -53,17 +53,7 @@
...
-
-							<image src="/assets/images/govuk-logotype-crown.png" class="govuk-header__logotype-crown-fallback-image"></image>
-						</svg>
-						<span class="govuk-header__logotype-text">
-							GOV.UK
-						</span>
-					</span>
+					LITE Internal
```
andytaylorkainos: i know you werent here, but we had a big conversation last week about how we HAVE to use lite-strings now - so can you make this a string?

PR #188, opened 2019-10-21
feat:LT-1566 server header removal

mh-hmcts:


docker-compose.yml
```diff
@@ -26,7 +26,9 @@ services:
       - 8080
     networks:
       - lite-api_default
-    command: pipenv run ./manage.py runserver 0.0.0.0:8080
+    command: pipenv run gunicorn -c gconfig.py conf.wsgi
+#    use this command instead to use django development server
+#    command: pipenv run ./manage.py runserver 0.0.0.0:8080
```
craiga: Is `docker-compose` only for use in local development? If so, I don't think we want to change this.

PR #196, opened 2019-10-22
Raise an explicit error about a string not existing in strings.json

craiga:
Adds a `StringNotFoundError`, which is raised when a string doesn't exist. The message includes the string which is missing.

### DIFF
--- a/core/builtins/custom_tags.py
+++ b/core/builtins/custom_tags.py
@@ -14,6 +14,11 @@ from core import strings
 register = template.Library()
 
 
+class StringNotFoundError(Exception):
+    def __init__(self, key):
+        super().__init__(f"String not found for key \"{key}\"")
+
+
 @register.simple_tag
 def get_string(value):
     """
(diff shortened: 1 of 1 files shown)


core/builtins/custom_tags.py
```diff
@@ -26,12 +31,17 @@ def get_string(value):
...
             strings.constants = json.load(json_file)
 
-    def get(d, keys):
+    def get(d, keys, original_string=None):
+        if not original_string:
+            original_string = keys
         if "." in keys:
             key, rest = keys.split(".", 1)
-            return get(d[key], rest)
+            return get(d[key], rest, original_string)
```
mh-hmcts: do we need to guard against a key error here?

core/builtins/custom_tags.py
```diff
@@ -26,12 +31,17 @@ def get_string(value):
...
         else:
-            return d[keys]
+            try:
+                return d[keys]
```
mh-hmcts: maybe a couple of simple unit tests for this?

core/builtins/custom_tags.py
```diff
@@ -14,6 +14,11 @@
 register = template.Library()
 
 
+class StringNotFoundError(Exception):
+    def __init__(self, key):
+        super().__init__(f"String not found for key \"{key}\"")
```
andytaylorkainos: single quotations please

PR #198, opened 2019-10-24
Bug: LT-1487 Back queue

Mark-Spiers-kainos:


cases/views/goods.py
```diff
@@ -32,29 +37,30 @@ def get(self, request, **kwargs):
             params = dict()
             params['goods'] = request.GET.getlist('goods')
             params['level'] = 'goods'
-            post_url = '?' + convert_dict_to_query_params(params)
-            return redirect(reverse_lazy('cases:assign_flags', kwargs={'pk': case_id}) + post_url)
+            post_url = '&' + convert_dict_to_query_params(params)
```
BenAndrew-Kainos: When GET variable builder method is added we shouldn't need '&' and '?' specified in views anymore

cases/views_helpers.py
```diff
@@ -32,7 +33,8 @@ def get_case_advice(get_advice, request, case, user_team_final, team=None):
         'case': case,
         'title': case.get('application').get('name'),
         'all_advice': advice['advice'],
-        'permissions': permissions
+        'permissions': permissions,
+        'queue_params': '?queues=' + request.GET.get('queue', DEFAULT_QUEUE_ID)
```
BenAndrew-Kainos: Similar issue with us building the GET url's manually

conf/decorators.py
```diff
@@ -19,3 +20,17 @@ def inner_decorator(request, *args, **kwargs):
         return wraps(func)(inner_decorator)
 
     return decorator
+
+
+def process_queue_params():
+    def decorator(func):
+        def inner(request, *args, **kwargs):
+            queue_id = request.request.GET.get('queue', DEFAULT_QUEUE_ID)
```
BenAndrew-Kainos: `request.request`? Not sure what that decorator argument is. I think that this just needs one argument that would be `view_object` and so you would just use `request.view_object`

templates/cases/case/views/case-flags.html
```diff
@@ -4,7 +4,7 @@
             {{ flag.name }}
         </div>
     {% endfor %}
-	<a href="{% url 'cases:assign_flags' case.id %}?level=cases&items={{ case.id }}" id="application-edit-case-flags" class="app-flag app-flag--action">
+	<a href="{% url 'cases:assign_flags' case.id %}{{ queue_params }}&level=cases&items={{ case.id }}" id="application-edit-case-flags" class="app-flag app-flag--action">
```
BenAndrew-Kainos: Can we prebuild all the params and pass that instead please

templates/cases/case/views/goods-flags.html
```diff
@@ -5,7 +5,7 @@
 		</div>
 	{% endfor %}
 	{% if not hide_creation_link %}
-		<a class="app-flag app-flag--action" href="{% url 'cases:assign_flags' case.id %}?level=goods&items={{ good.id }}&origin={{ origin }}">
+		<a class="app-flag app-flag--action" href="{% url 'cases:assign_flags' case.id %}{{ queue_params }}&level=goods&items={{ good.id }}&origin={{ origin }}">
```
BenAndrew-Kainos: Same as above and all other examples

templates/cases/index.html
```diff
@@ -147,13 +147,13 @@
                 <div>
                     {% if 'query' in case.type.key %}
                         <div>
-                            <a class="govuk-link govuk-!-font-size-19 govuk-link--no-visited-state govuk-!-font-weight-bold" href="{% url 'cases:case' case.id %}?return_to={{ current_queue.id }}" style="text-decoration: none; margin-bottom: 5px!important; display: inline-block;">
+                            <a class="govuk-link govuk-!-font-size-19 govuk-link--no-visited-state govuk-!-font-weight-bold" href="{% url 'cases:case' case.id %}?queue={{ current_queue.id }}" style="text-decoration: none; margin-bottom: 5px!important; display: inline-block;">
```
BenAndrew-Kainos: If possible modify existing decorator to also build these queue params

PR #203, opened 2019-10-28
Ensure paragraph ordering is maintained. Ensure multiple selections of case type are allowed.

craiga:


letter_templates/services.py
```diff
@@ -2,12 +2,20 @@
 from conf.constants import PICKLIST_URL, LETTER_TEMPLATES_URL
 
 
+def sort_letter_paragraphs(paragraphs, ids):
+    """Order a list of letter paragraphs in the same order as a list of IDs."""
+    paragraphs_by_id = {p["id"]: p for p in paragraphs}
+    return [paragraphs_by_id[id] for id in ids if id in paragraphs_by_id]
```
craiga: Still need to fix this up as per a previous conversation with @BenAndrew-Kainos

Pipfile
```diff
@@ -28,14 +28,14 @@ requests-oauthlib = "*"
...
 prospector = "*"
-lite-forms = "*"
 django-markdownx = "*"
 weasyprint = "*"
 django-csp = "*"
 bandit = "*"
 bleach = "*"
 gunicorn = "*"
 whitenoise = "*"
+lite-forms = {editable = true,path = "./lite-forms"}
```
BenAndrew-Kainos: Why do we need this for a submodule?

.gitmodules
```diff
@@ -7,3 +7,6 @@
 [submodule "ui_automation_tests/shared"]
 	path = ui_automation_tests/shared
 	url = git@github.com:uktrade/lite-tests-common.git
+[submodule "lite-forms"]
+	path = lite-forms
```
BenAndrew-Kainos: We can't have `-` in python modules. I would do `lite_forms`

letter_templates/services.py
```diff
@@ -2,12 +2,20 @@
...
+
+
 def get_letter_paragraphs(request, ids: list):
     if not ids:
         return []
 
     data = get(request, PICKLIST_URL + '?type=letter_paragraph' + '&ids=' + ','.join(ids))
-    return data.json()['picklist_items']
+    letter_paragraphs = data.json()['picklist_items']
+    letter_paragraphs = sort_letter_paragraphs(letter_paragraphs, ids)
```
BenAndrew-Kainos: Can we put the sorting into the `get` function as this will always be needed

letter_templates/tests.py
```diff
@@ -1,13 +1,26 @@
...
 
 
 class LetterTemplateEditLetterParagraphsTestCase(TestCase):
+
+    def test_variable_highlight(self):
+        """
+        Ensure that variable_highlight returns the correct value
+        """
+        test_input = '{% if True %}{{ applicant.name }}{% endif %}' * 2
+        expected_output = (ALT_OPEN_TAG + '{% if True %}' + CLOSE_TAG + OPEN_TAG +
```
BenAndrew-Kainos: Could you leave comments to explain this structure. I'm a bit confused to what it produces

letter_templates/tests.py
```diff
@@ -18,13 +31,3 @@ def test_sort_letter_paragraphs(self):
         )
         result_ids = [r["id"] for r in result]
```
BenAndrew-Kainos: Can we use more descriptive variable names.

letter_templates/views/generator.py
```diff
@@ -16,7 +16,10 @@ def get(self, request, **kwargs):
         return form_page(request, add_letter_template().forms[0])
 
     def post(self, request, **kwargs):
-        response, _ = submit_paged_form(request, add_letter_template(), post_letter_templates)
+        response, _ = submit_paged_form(request,
```
BenAndrew-Kainos: I know this is common place in the system at the moment but could we drop the `_` and do `[0]` on the end of the function call instead

letter_templates/views/generator.py
```diff
@@ -81,7 +86,19 @@ def post(self, request):
...
-        letter_paragraphs = request.POST.getlist('letter_paragraphs')
-        post_letter_templates(request, request.POST, letter_paragraphs)
-        messages.success(request, 'The letter template was created successfully')
-        return redirect('letter_templates:letter_templates')
+        letter_template_data = request.POST.copy()
+        letter_template_data["letter_paragraphs"] = request.POST.getlist('letter_paragraphs')
+        letter_template_data["restricted_to"] = request.POST.getlist('restricted_to')
+        response, status = post_letter_templates(request, letter_template_data)
+
+        if 200 <= status < 300:
```
BenAndrew-Kainos: Can we make this a check for the exact status code we expect (`201` I think)

letter_templates/views/generator.py
```diff
@@ -81,7 +86,19 @@ def post(self, request):
...
+            messages.success(request, 'The letter template was created successfully')
```
BenAndrew-Kainos: This message should be moved to `strings.json`

letter_templates/views/generator.py
```diff
@@ -81,7 +86,19 @@ def post(self, request):
...
+            return redirect('letter_templates:letter_templates')
+        else:
+            error_messages = []
+            errors = response["errors"]
+            for field, field_errors in errors.items():
+                for field_error in field_errors:
```
BenAndrew-Kainos: You can change `for field, field_errors in errors.items():` to `for field_errors in errors.values()` as you don't use the key

letter_templates/views/generator.py
```diff
@@ -81,7 +86,19 @@ def post(self, request):
...
+                    error_messages.append(field_error)
```
BenAndrew-Kainos: Also are we sure the API returns a list of errors for each field and not just a single item?

templates/letter_templates/letter_paragraphs.html
```diff
@@ -17,7 +17,7 @@ <h1 class="govuk-fieldset__heading">
 
 			{% hidden_field 'name' name %}
             {% for restricted in restricted_to %}
-			    {% hidden_field 'restricted_to' restricted_to %}
+			    {% hidden_field 'restricted_to' restricted %}
```
BenAndrew-Kainos: Could you rename `restricted` to `case_type` please

templates/letter_templates/letter_templates.html
```diff
@@ -37,7 +37,7 @@ <h2 class="govuk-heading-l">{% block title %}Letter templates{% endblock %}</h2>
 							{{ letter_template.layout.name }}
 						</td>
 						<td class="govuk-table__cell">
-							{% for restriction in letter_template.restricted_to %}{% if not forloop.first %}{% if forloop.last %} and {% else %}, {% endif %}{% endif %}{{ restriction.value }}{% endfor %}
+							{% for restriction in letter_template.restricted_to_display %}{% if not forloop.first %}{% if forloop.last %} and {% else %}, {% endif %}{% endif %}{{ restriction }}{% endfor %}
```
BenAndrew-Kainos: What is `restricted_to_display`? Also rename `restricted` as above

Pipfile
craiga: I tried to just import lite forms as a Python submodule, but it required a lot of changes which broke tests … this was the lightest-touch option I could figure out.

letter_templates/services.py
craiga: Not sure what you mean here … this _is_ the get function, isn't it?

letter_templates/tests.py
craiga: I don't understand it myself TBH. It's not something I've added; `git`'s just mucking up the diff a little. I'll try switching order of these test cases which might make things easier for it.

letter_templates/tests.py
craiga: What would you suggest?

templates/letter_templates/letter_templates.html
craiga: It's a list of display names for the `restricted_to` field.

In Django models, every field with choices gets a `get_FOO_display` method. The `_display` prefix is generally understood to be a user-readable version of some underlying value.

craiga:
@BenAndrew-Kainos Regarding renaming `restricted_to` to `case_type`, can we do this as part of a subsequent set of PRs? This will require API and frontend changes.

Pipfile
BenAndrew-Kainos: Yeah might be to do with the fact the submodule name has a `-` in it. If the submodule is named `lite_forms` the imports should be the same (`from lite_forms import ...` should work the same)

.gitmodules
BenAndrew-Kainos: Yeah we need to remove `lite-forms` as an install so that it's no longer confused.

letter_templates/tests.py
BenAndrew-Kainos: `[paragraph["id"] for paragraph in sorted_letter_paragraphs]` and rename `result` to `sorted_letter_paragraphs`

BenAndrew-Kainos: [paragraph["id"] for paragraph in sorted_letter_paragraphs] and rename result to sorted_letter_paragraphs

Dockerfile
```diff
@@ -12,5 +12,6 @@ RUN apt-get install -y libpq-dev gcc \
 
 RUN pip3 install pipenv
 ADD Pipfile* /app/
+ADD lite-forms /app/lite-forms
```
BenAndrew-Kainos: Do we need this with submodules?

.circleci/config.yml
```diff
@@ -68,7 +68,7 @@ jobs:
       # Linting: dodgy, mccabe, pep8, profile-validator, pyflakes, pylint
       - run:
           name: Run Linting
-          command: pipenv run prospector
+          command: pipenv run prospector --ignore-paths lite-forms
```
BenAndrew-Kainos: Is this needed because lite-forms doesn't pass pep8?

letter_templates/services.py
```diff
@@ -2,12 +2,23 @@
 from conf.constants import PICKLIST_URL, LETTER_TEMPLATES_URL
 
 
+def sort_letter_paragraphs(paragraphs, ids):
+    """Order a list of letter paragraphs in the same order as a list of IDs."""
+    for id in ids:
+        for paragraph in paragraphs:
+            if id == paragraph["id"]:
+                yield paragraph
```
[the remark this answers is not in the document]
BenAndrew-Kainos: Or just appending to a list

letter_templates/tests.py
BenAndrew-Kainos: Still needs renaming as above

letter_templates/views/create.py
```diff
@@ -18,7 +17,10 @@ def get(self, request, **kwargs):
 
     @staticmethod
     def post(request):
-        response = submit_paged_form(request, add_letter_template(), post_letter_template)[0]
+        response, _ = submit_paged_form(request,
```
BenAndrew-Kainos: Would prefer `response = submit_paged_form(request, add_letter_template(), post_letter_template)[0]`

letter_templates/views/letter_paragraphs.py
```diff
@@ -39,7 +53,9 @@ def _remove_letter_paragraph(template_content):
     def post(self, request):
         template_content = get_template_content(request)
         if template_content['action'].lower() == 'add_letter_paragraph':
-            return self._add_letter_paragraph(request, template_content)
+            return self._preview(request, template_content)
```
BenAndrew-Kainos: Why is `self._preview(request, template_content)` returned when the action is `add_letter_paragraph`

letter_templates/services.py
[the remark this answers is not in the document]
craiga: I don't care, but a list comprehension _is_ a generator (i.e. it uses `yield`).

letter_templates/views/generator.py
craiga: Why? A 2xx code signals success, and that's what we care about. Checking for 201 specifically means we can't handle any other kind of success the API may return in the future.

letter_templates/views/generator.py
craiga: Yeah, but then I'd need to unpack that `field_errors` tuple.

IMO it's easier to read this than having to include something like `for field_error in field_errors[1]`.

letter_templates/views/generator.py
craiga: Yep.

```python
return JsonResponse(data={'errors': serializer.errors}, status=status.HTTP_400_BAD_REQUEST)
```

templates/letter_templates/letter_paragraphs.html
craiga: Can you bring this up in https://github.com/uktrade/lite-api/pull/146 ? `restricted_to` is a field added as part of the bigger feature and I don't want to bloat this PR anymore than it already has been.

letter_templates/services.py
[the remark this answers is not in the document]
craiga: Also, I had this as a list comprehension and you told me to change it :)

letter_templates/views/create.py
craiga: Why?

letter_templates/views/generator.py
BenAndrew-Kainos: But this specific operation (a POST creating a template) should always be a `201`. If for some reason the same endpoint was used for some other kind of interaction I'd agree, but it shouldn't be

letter_templates/views/generator.py
BenAndrew-Kainos: No I mean for each field. serializer.errors as far as I'm aware is formatted like `{"field_name": "error_msg", ...}` and so doesn't display multiple errors for the same field, meaning you wouldn't have to iterate over it.

letter_templates/services.py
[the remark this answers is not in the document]
BenAndrew-Kainos: Actually it was a dictionary & list comprehension but don't worry. I've put a suggested change in the other branch

letter_templates/views/create.py
BenAndrew-Kainos: As `submit_paged_form` returns `response, status_code` and we only need the first item in this case `[0]` would do the same job. We do use `_` in a lot of places in the system but we've started changing over to this

BenAndrew-Kainos:
Happy to merge once changes in `LT-1029-refactor-ordering-paragraphs` are pulled in :)

PR #208, opened 2019-10-30
feat: LT-982 - Audit anchor

andytaylorkainos:


conf/constants.py
```diff
@@ -37,6 +37,8 @@
 PICKLIST_URL = '/picklist/'
 GOOD_CLC_REVIEW_URL = '/goods/controlcode/'
 MANAGE_STATUS_URL = '/status/'
+USERS_URL = '/users/'
```
craiga: This doesn't appear to be used anywhere other than the line below.

templates/cases/case/application-case.html
```diff
@@ -12,6 +12,10 @@
 
 {% block body %}
 
+<a id="case-notification" class="lite-info-bar app-info-bar--link lite-info-bar--no-animation" href="#case-activity-{{ notification.case_activity }}">
```
craiga: Why `app-info-bar--link` and not `lite-info-bar--link`?

andytaylorkainos: Jan helped me with this; its because one of the classes exists inside lite-content whilst the other exists inside the application

conf/constants.py
andytaylorkainos: Removed the variable :)

PR #213, opened 2019-11-04
Get and use letter layouts from the API. Also use new UUID IDs.

craiga:


letter_templates/views/create.py
```diff
@@ -36,12 +36,17 @@ def post(request):
...
             messages.success(request, get_string('letter_templates.letter_templates.successfully_created_banner'))
+
         else:
             error_messages = []
             errors = response["errors"]
-            for field_error in errors.values():
-                error_messages.append(field_error)
+            for field_errors in errors.values():
+                for field_error in field_errors:
+                    error_messages.append(field_error)
```
craiga: I found that `errors.values()` is a list of lists.

letter_templates/helpers.py
```diff
@@ -29,9 +30,16 @@ def generate_preview(layout, letter_paragraphs: list):
 
 def get_template_content(request):
     data = request.POST.copy()
+
+    layout = None
+    if data.get('layout'):
+        layout, status = get_letter_layout(request, data['layout'])
+        if status != 200:
+            raise RuntimeError(f"Letter layout endpoint returned { status }.")
```
BenAndrew-Kainos: How is this handled on the frontend (does it display a yellow screen?)

craiga: Yep, I'll go through to the normal "500" error handler. I think it's an appropriate response to an unknown letter layout ID being posted.

BenAndrew-Kainos: sure

craiga: The yellow screens only appear with `DEBUG = True` in settings.py.

There's a `500.html` which is displayed when an unhandled error bubbles up to the top.

craiga: Specific error message banners and calls to `error_page` should only really be used when there's an error the user is likely to encounter and we need to advise them on how to resolve it.

This isn't something the user is likely to encounter.

BenAndrew-Kainos: Although I understand your argument, currently we always create our own error page. Could you raise this briefly after standup

BenAndrew-Kainos: As discussed, looks good :)

PR #221, opened 2019-11-11
Task: context_variables unit testing

BenAndrew-Kainos:


BenAndrew-Kainos:
> Whats the idea behind these context variables?

They are used in letter templates to prompt internal users when they type '{' with a list of possible variables for them to add to the paragraph

letter_templates/tests.py
```diff
@@ -31,3 +33,50 @@ def test_variable_highlight(self):
...
+
+        from letter_templates.context_variables import context_variables, flattened_context_variables, JSON_PATH
+
+        self.context_variables = context_variables
+        self.flattened_context_variables = flattened_context_variables
+        with open(JSON_PATH, 'r') as f:
+            self.variables = json.load(f)
+
+    def _validate_variable_dict(self, dictionary):
+        for key, value in dictionary.items():
```
stephanieboyd: Add docstrings to these functions, please

letter_templates/tests.py
```diff
@@ -31,3 +33,50 @@ def test_variable_highlight(self):
...
+        if "letter_templates" in os.getcwd():
+            os.chdir('../')
+            sys.path.insert(0, os.getcwd())
+
+        from letter_templates.context_variables import context_variables, flattened_context_variables, JSON_PATH
+
+        self.context_variables = context_variables
+        self.flattened_context_variables = flattened_context_variables
+        with open(JSON_PATH, 'r') as f:
+            self.variables = json.load(f)
```
stephanieboyd: Can we name this something better? `self.variables` doens't read well

letter_templates/tests.py
```diff
@@ -31,3 +33,62 @@ def test_variable_highlight(self):
...
 
         self.assertEqual(variable_highlight(test_input), expected_output)
+
+
+class ContextVariablesTestCase(TestCase):
+    def setUp(self):
+        # Parent directory appended to access JSON file in lite-content
+        if "letter_templates" in os.getcwd():
+            os.chdir('../')
+            sys.path.insert(0, os.getcwd())
```
craiga: This seems dangerous! Why do we need to do this?

cases/views/clc_query.py
```diff
@@ -58,7 +58,7 @@ def post(self, request, **kwargs):
             response, response_data = submit_single_form(request,
                                                          self.form,
                                                          put_control_list_classification_query,
-                                                         object_pk=str(self.case['query']['id']),
+                                                         pk=str(self.case['query']['id']),
```
craiga: Why is this changing in this PR? Seems unrelated.

cases/views/main.py
```diff
@@ -194,7 +194,7 @@ def post(self, request, **kwargs):
         response, data = submit_single_form(request,
                                             move_case_form(request, reverse('cases:case', kwargs={'pk': case_id})),
                                             put_case,
-                                            object_pk=case_id,
+                                            pk=case_id,
```
craiga: Why is this changing in this PR? Seems unrelated.

BenAndrew-Kainos: It is. But for some reason someone has updated lite-forms but not updated these references so CircleCI fails :(

letter_templates/tests.py
BenAndrew-Kainos: This is for the unit test only. The load context_variables class opens the JSON through a relative path and the tests are in a directory further in so I had to cd to get it to work correctly. If you have a minute I would love to improve this too

BenAndrew-Kainos: All sorted :)

letter_templates/context_variables.py
```diff
@@ -77,6 +79,7 @@ def flatten_dict(dictionary, path):
 
 context_variables = load_context_variables()
```
craiga: Oops, wish I noticed this before.

Every time anyone imports anything from this module, `load_context_variables` will be called. This is potentially a problem because:

1. it slows things down; and
2. it will raise an error if whatever imports this isn't able to read the JSON file from disk.

Ideally, `load_context_variables` should be called directly by whatever needs it.

It's not something you've changed in this PR so I won't hold it back, but this should be fixed at some point.

The same goes for `flattened_context_variables`.

letter_templates/tests.py
```diff
@@ -31,3 +32,53 @@ def test_variable_highlight(self):
...
+        for key, value in self.context_variables_json.items():
+            if 'variables' in value:
+                self.variables_in_json[key] = value['variables']
+            else:
+                self.variables_in_json[key] = {}
+
+        self._validate_variable_dict(context_variables)
+
+    def test_flatten_context_variables(self):
+        self._validate_flattened_variables(context_variables, '')
```
craiga: A better unit test for `flatten_context_variables` would be something like this:

```
flattened = flatten_context_variables({
    "no_children": "Quark",
    "with_children": {"brother": "Rom"}
})
self.assertEquals("Quark", flattened["no_children"])
self.assertEquals("Rom", flattened["with_children.brother"])
```

letter_templates/tests.py
```diff
@@ -31,3 +32,53 @@ def test_variable_highlight(self):
                            '{{ applicant.name }}' + CLOSE_TAG + ALT_OPEN_TAG + '{% endif %}' + CLOSE_TAG) * 2
 
         self.assertEqual(variable_highlight(test_input), expected_output)
+
+
+class ContextVariablesTestCase(TestCase):
+    def setUp(self):
+        with open(JSON_PATH, 'r') as f:
+            self.context_variables_json = json.load(f)
```
craiga: Strictly, unit tests shouldn't have any external dependencies. This includes the filesystem.

letter_templates/tests.py
BenAndrew-Kainos: Unfortunately there are roughly 130 variables produced and this are likely to change significantly over time so I decided not to hard code a list of variables to check (which would also defeat the purpose of these methods anyway)

letter_templates/context_variables.py
BenAndrew-Kainos: You are 100% correct. Should have thought of that earlier. my bad.

PR #225, opened 2019-11-12
Bug: Letter Template CSS

BenAndrew-Kainos:


assets/css/letter_templates/_ecju.scss
```diff
@@ -0,0 +1,65 @@
+.ecju-template{
```
craiga: In most other places there's a space before the`{`, but if it compiles it's no problem.

letter_templates/layouts/ecju.html
```diff
@@ -1,63 +1,62 @@
...
-      <p>
-        3 Whitehall Place<br>
-        London<br>
-        SW1A 2AW
-      </p>
-      <p>
-        Tel +44 (0)20 7215 4594<br>
-        Enquiries +44 (0)20 7215 5000
+<div class="ecju-template">
+    <body>
```
craiga: Would probably be better to apply the class to the `body` element or a direct descendant of it. However, we can fix that up in the next story.

PR #236, opened 2019-11-15
feat: LT-1694 back to queue link

Mark-Spiers-kainos:


cases/views/main.py
```diff
@@ -75,14 +75,12 @@ def get(self, request, **kwargs):
         queue_name = request.GET.get('queue_name')
 
         case['all_flags'] = _get_all_distinct_flags(case)
-        total_goods_value = _get_total_goods_value(case)
```
Mark-Spiers-kainos: this is a fix for not being able to see clc and euas currently

PR #237, opened 2019-11-15
hotfix: Total Goods Value on HMRC Query 

cstobart:


cases/views/main.py
```diff
@@ -93,7 +93,6 @@ def get(self, request, **kwargs):
             context['good'] = case['query']['good']
             return render(request, 'cases/case/queries/clc-query-case.html', context)
         elif case.get('application').get('application_type').get('key') == 'hmrc_query':
```
mh-hmcts: can we use constants instead of hardcoded strings please 'hmrc_query'

PR #244, opened 2019-11-21
Refactor: Case inheritance

Mark-Spiers-kainos:


templates/case/standard-licence-case.html
```diff
@@ -112,7 +112,7 @@ <h2 class="govuk-heading-l">
 			</div>
 			<div class="govuk-grid-column-one-quarter">
 				<p class="app-case-board__heading">{% lcs 'CASE_INFO_LAST_UPDATED' %}</p>
-				<p>{{ case.application.last_modified_at|str_date }}</p>
+				<p>{{ case.application.modified|str_date }}</p>
```
andytaylorkainos: same as question on exporter, howcome its not `_at`?

PR #246, opened 2019-11-22
hotfix: Fixed CaseType enum implementation & one other bug

efahy:


core/builtins/custom_tags.py
```diff
@@ -227,7 +227,7 @@ def get_address(data):
         address["postcode"],
         address["country"]["name"],
     ]
-    return ", ".join([x for x in address if len(x) is not 0 and x is not None])
+    return ", ".join([x for x in address if x])
```
stephanieboyd: Much improved

cases/views/main.py
```diff
@@ -173,9 +172,9 @@ def get(self, request, **kwargs):
             ]
         }
 
-        if case["type"]["key"] == CaseType.APPLICATION:
+        if case["type"]["key"] == CaseType.APPLICATION.value:
```
stephanieboyd: Pull this out into a case_type variable, keeps it consistent with the other view and lookup happens once

efahy: Done

PR #249, opened 2019-11-27
feat: LT-1671 - Permissions to create and edit document templates

jfaracik:


conf/constants.py
```diff
@@ -54,3 +54,12 @@
 
 # ROLE IDS
 SUPER_USER_ROLE_ID = "00000000-0000-0000-0000-000000000002"
+
+
+class Permissions:
```
a-bek: Shall we change this to an Enum class?

You'll have to update the signature of `def has_permission(permission: str):`

Also must update where you use it, - `if helpers.has_permission(request, permission):` to `permission.value`

conf/decorators.py
```diff
@@ -1,21 +1,26 @@
...
+        @wraps(view_func)
+        def _wrapped_view(request, *args, **kwargs):
+            if helpers.has_permission(request, permission):
+                return view_func(request, *args, **kwargs)
 
-            raise Http404()
+            raise PermissionDeniedError(
+                f"You don't have the permission '{permission}' to view this, "
+                "check urlpatterns or the function decorator if you want to change "
+                "this functionality."
```
markm5A8Z9F43913A: Content?

jfaracik: This is solely for devs, regular users won't see this :)

conf/constants.py
jfaracik: Good suggestion! I've done it :)

core/helpers.py
```diff
@@ -6,4 +11,29 @@ def convert_dict_to_query_params(dictionary):
...
+
+def has_permission(request, permission: Permission):
+    """
+    Returns true if the user has a given permission, else false
+    """
+    user_permissions = get_user_permissions(request)
+    return permission.value in user_permissions
+
+
+def decorate_patterns_with_permission(patterns, permission: Permission):
```
a-bek: 👍

PR #251, opened 2019-12-02
feat: LT-1622 - Restrict Roles and Permissions Visible to User

cstobart:


users/forms/users.py
```diff
@@ -21,14 +22,15 @@ def add_user_form(request):
     )
 
 
-def edit_user_form(request, user_id, super_user):
+def edit_user_form(request, user_id, cannot_edit_role: Boolean):
```
Mark-Spiers-kainos: was wondering if you could just use bool? and not import

users/forms/users.py
```diff
@@ -1,4 +1,5 @@
 from django.urls import reverse_lazy
+from setoptconf import Boolean
```
stephanieboyd: Do we already install this package?

users/forms/users.py
cstobart: good point

users/forms/users.py
cstobart: good point

users/forms/users.py
```diff
@@ -21,14 +20,15 @@ def add_user_form(request):
     )
 
 
-def edit_user_form(request, user_id, super_user):
+def edit_user_form(request, user_id, cannot_edit_role: bool):
     return Form(
         title="Edit User",
         questions=[
             TextInput(title="Email", name="email"),
             Select(name="team", title="What team will the user belong to?", options=get_teams(request, True)),
```
stephanieboyd: In the interest of tidying as we go - move to strings? Sorry for not spotting it the first time!

users/forms/users.py
```diff
@@ -21,14 +20,15 @@ def add_user_form(request):
     )
 
 
-def edit_user_form(request, user_id, super_user):
+def edit_user_form(request, user_id, cannot_edit_role: bool):
```
andytaylorkainos: can we invert the boolean please? (`can_edit_role`) and drop the `not` in the line 30 below

users/views/users.py
```diff
@@ -65,17 +65,19 @@ class EditUser(TemplateView):
     def get(self, request, **kwargs):
         user, _ = get_gov_user(request, str(kwargs["pk"]))
         super_user = is_super_user(user) and request.user.lite_api_user_id == str(kwargs["pk"])
-        return form_page(request, edit_user_form(request, str(kwargs["pk"]), super_user), data=user["user"])
+        cannot_edit_role = user["user"]["id"] == request.user.lite_api_user_id or super_user
```
andytaylorkainos: please invert the boolean

cstobart: done

users/views/users.py
```diff
@@ -65,7 +65,7 @@ class EditUser(TemplateView):
     def get(self, request, **kwargs):
         user, _ = get_gov_user(request, str(kwargs["pk"]))
         super_user = is_super_user(user) and request.user.lite_api_user_id == str(kwargs["pk"])
-        can_edit_role = (user["user"]["id"] != request.user.lite_api_user_id) or not super_user
+        can_edit_role = not (user["user"]["id"] == request.user.lite_api_user_id or super_user)
```
andytaylorkainos: `can_edit_role = user["user"]["id"] != request.user.lite_api_user_id`

PR #254, opened 2019-12-02
Fix, tags, and warnings

phlptt:
* Fix to title missing on CLC queries
* Getting rid of a few warnings from deprecated stuff
* Added a few tags to introduce smoke tests and regression tests
** Smoke tags are used to run a small subset of tests, ie key functionality so that it can be quicker to run through them all. 
** Regression tags are added to tests which aren't smoke tags so that when all smoke tags have been run you can run @regression tag to run the rest of them.
** All tags are added so there's a tag for all of them but this isn't necessary.
https://www.bytelion.com/smoke-testing-vs-regression-testing/


ui_automation_tests/features/add_goods.feature
```diff
@@ -4,7 +4,7 @@ Feature: I want to add a clc-case good to the goods list
     I want to add a clc-case good to the goods list
     So that I can ensure the good is listed in my cases
 
-    @LT_1006_add_clc_query_good
+    @LT_1006_add_clc_query_good @regression
```
BenAndrew-Kainos: What's @regression for?

ui_automation_tests/features/assign_case_flags_to_case.feature
```diff
@@ -32,7 +32,7 @@ Feature: I want to add case-level flags to a case and view them
     Then I see previously created application
     And I see the added flags on the queue
 
-  @LT_951
+  @LT_951 @smoke
```
BenAndrew-Kainos: What's @smoke for?

ui_automation_tests/step_defs/test_case_notes.py
```diff
@@ -75,5 +75,5 @@ def click_visible_to_exporters_checkbox(driver):
 
 @when("I click confirm on confirmation box")
 def click_confirm_on_confirmation_box(driver):
-    alert = driver.switch_to_alert()
+    alert = driver.switch_to.alert
```
BenAndrew-Kainos: Is switch to alert not a method call?

templates/case/queries/clc-query-case.html
```diff
@@ -19,7 +19,7 @@
 	<div class="lite-app-bar">
 		<div class="lite-app-bar__content">
 			<p class="govuk-caption-m" style="margin: 0;">{{ case.type.value }} by <a class="govuk-link govuk-link--no-visited-state" style="display: inline-block;" href="{% url 'organisations:organisation' case.query.organisation.id %}">{{ case.query.organisation.name }}</a></p>
-			<h2 class="govuk-heading-l">{{ case.query.id|reference_code }}</h2>
+			<h2 class="govuk-heading-l">{% block title %}{{ case.query.id|reference_code }}{% endblock %}</h2>
```
andytaylorkainos: I think this will be fixed by https://github.com/uktrade/lite-internal-frontend/pull/238/files#diff-613cf1165c633cc070e68194baf2bf14 (LT-1028, which is ahead of this in the merge queue)

ui_automation_tests/step_defs/test_case_notes.py
phlptt: yes but it seems to work

ui_automation_tests/features/add_goods.feature
phlptt: uodated the descrpiptuon

templates/case/queries/clc-query-case.html
phlptt: o.

ui_automation_tests/step_defs/test_case_notes.py
phlptt: 14:57:51  ui_automation_tests/step_defs/test_case_notes.py::test_add_a_new_exporter_visible_case_note
14:57:51    /home/jenkins/agent/workspace/lite-e2e-internal-frontend/ui_automation_tests/step_defs/test_case_notes.py:78: DeprecationWarning: use driver.switch_to.alert instead
14:57:51      alert = driver.switch_to_alert()

BenAndrew-Kainos: ok

PR #258, opened 2019-12-03
test: Lt 1807 hmrc case view

phlptt:
Note i haven't written it well as this will get rewritten when a bug is fixed


ui_automation_tests/step_defs/test_hmrc_query.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@given("I create a hmrc query")  # noqa
+def create_hmrc_query(driver, apply_for_hmrc_query, context):
+    pass
+
+
+@when("I go to HMRC query")  # noqa
+def go_to_hmrc(driver):
+    driver.find_element_by_css_selector('.govuk-link[href*="cases"]').click()
```
BenAndrew-Kainos: Should this not be in a Page class

ui_automation_tests/step_defs/test_hmrc_query.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@then("I see HMRC query")
+def see_hmrc(driver):
+    assert driver.find_element_by_css_selector(".govuk-caption-m").text == "HMRC Query"
```
BenAndrew-Kainos: Same as above

PR #266, opened 2019-12-05
chore: SSO refactor

jfaracik:
* Remove unused dependencies and files
* Bring authentication under one roof: `auth`
* Remove first and last name from organisation sign up
* Add "first and last name" field for individual organisation


organisations/forms.py
```diff
@@ -118,11 +113,7 @@ def register_hmrc_organisation_forms(name=""):
             ),
             Form(
                 title="Create an admin for " + name,
```
Mark-Spiers-kainos: strings

ui_automation_tests/pages/organisations_form_page.py
```diff
@@ -49,15 +46,8 @@ def enter_country(self, text):
         country_tb = self.driver.find_element_by_id("site.address.country")
         country_tb.send_keys(text)
 
+    def enter_first_last_name(self, text):
```
efahy: Can you rename to make it clear(er) which is for org and which for user, e.g. `enter_org_name` and `enter_user_first_last_name`

PR #271, opened 2019-12-10
test: Speeding up tests

phlptt:
🥳 8m 04s for all the internal tests..
* Using open applications instead of standard where possible
* Changes via API with how the document is seeded ( no scan) - there are still timeouts of a document not processing however so this is not fully resolved
* Changed the timeout to 0 for when its looking that something doesnt exist
* Tidied up the file names so they mapped more logically
* Bunched a test together to make it quicker (less set up)


ui_automation_tests/step_defs/test_assign_users_to_queue.py
```diff
@@ -2,11 +2,11 @@
 from pages.case_list_page import CaseListPage
 from pages.shared import Shared
 
-scenarios("../features/users_to_queue.feature", strict_gherkin=False)
+scenarios("../features/assign_users_to_queue.feature", strict_gherkin=False)
```
stephanieboyd: (Y)

PR #280, opened 2019-12-16
feat: LT-1186 - Destination Flags

cstobart:


templates/case/standard-licence-case.html
```diff
@@ -183,27 +183,42 @@ <h2 class="govuk-accordion__section-heading">
 				</h2>
 			</div>
 			<div id="accordion-default-content-3" class="govuk-accordion__section-content" aria-labelledby="accordion-default-heading-3">
-				<table class="govuk-table" id="table-entities">
+                <form action="{% url 'cases:assign_destination_flags' case.id %}" method="get">
+                <button id="button-edit-goods-flags" class="govuk-button" data-module="govuk-button">
+                    Destination Flags [temp]
+{#                    {% lcs 'CASES.ApplicationPage.EDIT_FLAGS' %}#}
```
Mark-Spiers-kainos: why is this commented?

cstobart: I need to do the strings still so that reminds me to go back and do it

cases/views/destinations.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class AssignDestinationFlags(TemplateView):
+    def get(self, request, **kwargs):
+        case_id = str(kwargs["pk"])
+
+        params = dict()
+        params["items"] = request.GET.getlist("destinations")
+        params["level"] = "destinations"
+        post_url = "?" + convert_dict_to_query_params(params)
```
andytaylorkainos: use an f-string please

cases/views/destinations.py
```diff
@@ -0,0 +1,16 @@
...
+from django.views.generic import TemplateView
+
+from core.helpers import convert_dict_to_query_params
+
+
+class AssignDestinationFlags(TemplateView):
+    def get(self, request, **kwargs):
+        case_id = str(kwargs["pk"])
+
+        params = dict()
```
andytaylorkainos: change to `{}` please

conf/constants.py
```diff
@@ -45,6 +45,7 @@
 GENERATED_DOCUMENTS_PREVIEW_URL = GENERATED_DOCUMENTS_URL + "preview/"
 PREVIEW_URL = "/preview/"
 GENERATE_PREVIEW_URL = "generate-preview/"
+DESTINATION_URL = CASE_URL + "destination/"
```
andytaylorkainos: will need changed based on my api comment

cases/views/destinations.py
jfaracik: How would an f string help here?

cases/services.py
```diff
@@ -358,3 +359,8 @@ def get_generated_document_preview(request, pk, tpk, text):
 def get_generated_document(request, pk, dpk):
     data = get(request, CASE_URL + pk + GENERATED_DOCUMENTS_URL + dpk + "/")
     return data.json(), data.status_code
+
+
+def get_destination(request, pk):
+    data = get(request, DESTINATION_URL + pk)
+    return data.json(), data.status_code
```
jfaracik: Don't return a status code, instead just return the json like so: data.json()["destination"]

templates/case/open-licence-case.html
```diff
@@ -143,7 +143,7 @@ <h2 class="govuk-accordion__section-heading">
 						<thead class="govuk-table__head">
 							<tr class="govuk-table__row">
 								<th scope="col" class="govuk-table__header">
-									<button type="button" name="button" class="lite-button-checkbox" title="Select all/Deselect all"></button>
+									<button id="select-all-goods" type="button" name="button" class="lite-button-checkbox" title="Select all/Deselect all"></button>
```
jfaracik: Rename this to link-select-all-goods

PR #282, opened 2019-12-17
test: New test to go through menu and refactor everywhere else going directly to URL

phlptt:
New test to go through menu and refactor everywhere else going directly to URL
Running them locally they all passed in approx 7 mins so a bit faster

### DIFF
--- a/templates/layouts/base.html
+++ b/templates/layouts/base.html
@@ -59,7 +59,7 @@
 
 			{% if user and user.lite_api_user_id %}
 				<div id="lite-user-menu">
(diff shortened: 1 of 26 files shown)


templates/layouts/base.html
```diff
@@ -59,7 +59,7 @@
 
 			{% if user and user.lite_api_user_id %}
 				<div id="lite-user-menu">
-					<a href="{% url 'auth:logout' %}" class="lite-user-menu-button">
+					<a href="{% url 'auth:logout' %}" class="lite-user-menu-button" id="log-out-button">
```
jfaracik: "link-logout"

phlptt: cheers

PR #286, opened 2019-12-19
Test clean up

phlptt:
* Get Authbroker URL from .env
* Remove unused packages
* Remove unused functions and variables


ui_automation_tests/pages/assign_flags_to_case.py
```diff
@@ -2,12 +2,6 @@
 
 
 class CaseFlagsPages(BasePage):
-    def assign_flags(self, context):
-        flags = self.driver.find_elements_by_css_selector("input.govuk-checkboxes__input")
-        flags[0].click()
-        context.assigned_flag = flags[0].get_attribute("id")
-
-    def select_flag(self, context, flag_id):
+    def select_flag(self, flag_id):
```
andytaylorkainos: it says that the parameter for this is `flag_id` - but each time you call this method you pass it `context.flag_name` which one is correct? :)

PR #304, opened 2020-01-09
bug: LT-1835 document generation on euas

efahy:


ui_automation_tests/step_defs/test_eua_queries.py
```diff
@@ -24,9 +24,10 @@ def dropdown_contains_correct_functionality(driver):
     assert application_page.get_ecju_queries_element()
     assert application_page.is_change_status_available()
     assert application_page.get_case_officer_element()
+    assert application_page.get_generate_document_element()
     # This tests that the expected elements are the only ones that appear, and that any new functionality added
     # should be tested if not tested elsewhere.
-    assert len(driver.find_elements_by_css_selector("div.lite-app-bar__controls>div>a")) == 5
+    assert len(driver.find_elements_by_css_selector("div.lite-app-bar__controls>div>a")) == 6
```
BenAndrew-Kainos: Many issues with this line 

- Driver actions should only be in pages not in step defs. (So in this case you should make a function that returns the length and keep the assertion here)
- Selectors should be defined as constants in the page class.
- Also this css selector is ridiculous and very fragile. (please use an ID or Name)

PR #305, opened 2020-01-09
feat: Lt-1629 - review open application good ratings

Mark-Spiers-kainos:


cases/forms/review_goods_clc.py
```diff
@@ -6,14 +6,19 @@
 from picklists.services import get_picklists
 
 
-def review_goods_clc_query_form(request, back_url):
+def review_goods_clc_query_form(request, back_url, goods_type):
```
jfaracik: Could you rename goods_type to is_goods_type

Mark-Spiers-kainos: done

templates/case/review-goods.html
```diff
@@ -38,30 +38,49 @@ <h2 class="govuk-heading-l">{% lcs 'Cases.ReviewGoodsSummary.HEADING' %}</h2>
...
+							{{ flag.name}} <br>
+						{% endfor %}
+					</td>
+				</tr>
+			{% endfor %}
+		{% else %}
+			{% for good in objects %}
+				<tr class="govuk-table__row">
+					<td class="govuk-table__cell">{{ good.good.description }}</td>
+					<td class="govuk-table__cell">{{ good.good.report_summary|default_if_none:"N/A" }}</td>
```
jfaracik: Is there a reason why this isn't using default_na?

templates/case/review-goods.html
```diff
@@ -38,30 +38,49 @@ <h2 class="govuk-heading-l">{% lcs 'Cases.ReviewGoodsSummary.HEADING' %}</h2>
...
+					<td class="govuk-table__cell">{{ good.good.is_good_controlled|capfirst}}</td>
```
jfaracik: Could this not use sentence_case? - or better yet have the api return a key/value pair?

cases/views/goods.py
```diff
@@ -39,13 +39,19 @@ def get(self, request, **kwargs):
...
         goods_postfix_url = "?" + convert_dict_to_query_params(parameters)
 
-        for good in case["application"]["goods"]:
-            if good["good"]["id"] in goods_pk_list:
-                goods.append(good)
+        if case["application"]["application_type"]["key"] == "standard_licence":
+            for good in case["application"]["goods"]:
+                if good["good"]["id"] in goods_pk_list:
+                    goods.append(good)
+        else:
```
andytaylorkainos:
```
key = "goods" if case["application"]["application_type"]["key"] == "standard_licence" else "goods_types"

for good in case["application"][key]:
      if good["id"] in goods_pk_list:
            goods.append(good)
``

File: cases/views/goods.py [hunk 5]
```diff
@@ -107,7 +121,11 @@ def post(self, request, *args, **kwargs):
         response = post_goods_control_code(request, self.case_id, form_data)
 
         if response.status_code == 400:
-            form = review_goods_clc_query_form(request, self.back_link)
+            case = get_case(request, self.case_id)
+            if case["application"]["application_type"]["key"] == "open_licence":
+                form = review_goods_clc_query_form(request, self.back_link, is_goods_type=True)
+            else:
+                form = review_goods_clc_query_form(request, self.back_link, is_goods_type=False)
```
andytaylorkainos:
```
is_goods_type = case["application"]["application_type"]["key"] == "open_licence"

form = review_goods_clc_query_form(request, self.back_link, is_goods_type=is_goods_type)
```

templates/case/review-goods.html
```diff
@@ -29,39 +29,92 @@ <h2 class="govuk-heading-l">{% lcs 'Cases.ReviewGoodsSummary.HEADING' %}</h2>
...
-                <td class="govuk-table__cell">{{ good.good.comment}}</td>
-                <td class="govuk-table__cell">{{ good.quantity }} {{ good.unit.value }}/£{{ good.value }}</td>
-                <td class="govuk-table__cell">
-                    {% for flag in good.good.flags %}
-                        {{ flag.name}} <br>
-                    {% endfor %}
-                </td>
-			</tr>
-		{% endfor %}
+		{% if application_type == "standard_licence" %}
```
andytaylorkainos: could you not just do this logic in the view to get a list of goods and pass that in via context to avoid the duplication on the template?

andytaylorkainos: could you not do this logic in the view to get a list of goods and pass that list in via the context to avoid duplicating html in the template?

cases/views/goods.py
Mark-Spiers-kainos: you helped me spot a bug before it appeared

cases/views/goods.py
Mark-Spiers-kainos: As discussed this won't be possible.

cases/forms/review_goods_clc.py
```diff
@@ -6,14 +6,19 @@
 from picklists.services import get_picklists
 
 
-def review_goods_clc_query_form(request, back_url):
+def review_goods_clc_query_form(request, back_url, is_goods_type):
+    if is_goods_type:
+        options = [Option(key="True", value="Yes", show_pane="pane_control_code"), Option(key="False", value="No")]
```
jfaracik: Could this be done with conditional instead? 

eg 

```
[Option(key=conditional(is_goods_type, "True", "yes), value="Yes", show_pane="pane_control_code")...
```

templates/case/hmrc-case.html
```diff
@@ -97,11 +97,15 @@ <h3 class="govuk-heading-m">{% lcs 'CASES.HMRCPage.DenialReasons.TITLE' %}</h3>
...
 
-<form action="{% url 'cases:assign_flags' case.id %}" method="get">
-    <input type="hidden" name="level" value="goods"/>
-    <button id="button-edit-goods-flags" value="goods" class="govuk-button" data-module="govuk-button">
-        {% lcs 'Cases.Case.EDIT_GOODS_FLAGS' %}
-    </button>
+<form action="{% url 'cases:review_goods' case.id %}" method="get">
+	{% if 'REVIEW_GOODS' in permissions and not status_is_terminal %}
+		<button id="button-review-goods" name="action" value="review-goods" class="govuk-button" data-module="govuk-button">
+			{% lcs 'CASES.HMRCPage.Good.REVIEW_GOODS' %}
```
jfaracik: Why is CASES in all caps?

cases/views/goods.py
```diff
@@ -107,7 +126,10 @@ def post(self, request, *args, **kwargs):
         response = post_goods_control_code(request, self.case_id, form_data)
 
         if response.status_code == 400:
-            form = review_goods_clc_query_form(request, self.back_link)
+            case = get_case(request, self.case_id)
+            is_goods_type = case["application"]["application_type"]["key"] != "standard_licence"
```
jfaracik: Is there a const we could use instead of standard_licence?

templates/case/hmrc-case.html
Mark-Spiers-kainos: cause it is in the `cases.py` file.

jfaracik: Could this be title case to be consistent with exporter?

PR #307, opened 2020-01-10
feat: LT-1239 - Incorporated application

jfaracik:
* Added table columns to indicate if standard/open goods are incorporated or not


templates/case/open-licence-case.html
```diff
@@ -159,6 +159,7 @@ <h2 class="govuk-accordion__section-heading">
 								<th scope="col" class="govuk-table__header govuk-!-width-one-half">{% lcs 'GOOD_DESCRIPTION' %}</th>
 								<th scope="col" class="govuk-table__header">{% lcs 'GOOD_CONTROLLED' %}</th>
 								<th scope="col" class="govuk-table__header">{% lcs 'GOOD_CONTROL_LIST_ENTRY' %}</th>
+								<th scope="col" class="govuk-table__header">{% lcs 'GOOD_INCORPORATED' %}</th>
```
andytaylorkainos: These strings dont follow the correct format. They should be inside their own class afaik

PR #309, opened 2020-01-13
bug: LT-1952 Update Third Party design for sub type & role

BenAndrew-Kainos:
Don't worry, most of the changes are just updates in lite_content


templates/case/includes/parties.html
```diff
@@ -0,0 +1,128 @@
+<table class="govuk-table" id="table-entities">
+    <thead class="govuk-table__head">
+        <tr class="govuk-table__row">
+            <th scope="col" class="govuk-table__header">
+                <button id="link-select-all-destinations" type="button" name="button" class="lite-button-checkbox" title="Select all/Deselect all"></button>
```
stephanieboyd: Is this title not a strings candidate?

templates/case/includes/parties.html
```diff
@@ -0,0 +1,128 @@
...
+            </th>
+            <th scope="col" class="govuk-table__header"></th>
+            <th scope="col" class="govuk-table__header govuk-!-width-one-third">{% lcs 'CASE_PARTIES_NAME' %}</th>
```
stephanieboyd: Shouldn't there be whitespace between the tag `>` and the content?

templates/case/includes/parties.html
BenAndrew-Kainos: Yep. Will move now

PR #311, opened 2020-01-14
feat 1358: Add statuses to roles

stephanieboyd:


core/builtins/custom_tags.py
```diff
@@ -248,6 +248,12 @@ def sentence_case(value):
     return " ".join([words[0].capitalize()] + words[1:])
 
 
+def lowercase_and_underscore(value):
+    """Change value to uppercase on initial word and preserve casing on all other words. """
```
Mark-Spiers-kainos: the string and function name kinda clash

stephanieboyd: copy paste error, fixed

core/services.py
```diff
@@ -58,10 +61,40 @@ def get_countries(request, convert_to_options=False):
...
 
 
+def get_statuses_as_options(request):
+    """ Get a list of statuses as Options. """
+    data, _ = get_statuses(request)
+    return [Option(key=item["id"], value=sentence_case(item["status"])) for item in data.get("statuses")]
+
+
+def get_permissible_statuses(request, case_id):
+    """ Get a list of case statuses permissible for the user's role. """
```
Mark-Spiers-kainos: would it be possible to pass the case_type in here rather than case_id?

cases/services.py
```diff
@@ -42,6 +42,7 @@ def put_case(request, pk, json):
 
 # Applications
 def put_application_status(request, pk, json):
+    json = format_status_in_request_data(json)
```
andytaylorkainos: howcome we have to copy the data in `format_status_in_request_data` if we are overwriting the original data on this line anyway?

stephanieboyd: Cause the original object is immutable, we're just reassigning it back into the name `json`.

cases/views/main.py
```diff
@@ -177,37 +170,22 @@ class ManageCase(TemplateView):
...
         case_id = str(kwargs["pk"])
         case = get_case(request, case_id)
-        statuses, _ = get_statuses(request)
-
-        reduced_statuses = {
-            "statuses": [
-                status for status in statuses["statuses"] if status["status"] in ["closed", "submitted", "withdrawn"]
-            ]
-        }
+        statuses = get_permissible_statuses(request, case_id)
```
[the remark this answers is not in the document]
stephanieboyd: Yeah, good suggestion. Changed :)

core/services.py
stephanieboyd: Good suggestion, thanks

PR #318, opened 2020-01-17
task: Django version

phlptt:


phlptt:
The reason the Circle CI test is failing is because this is what this PR is fixing! So, when this is merged then in the future that test will pass.

phlptt:
The reason the Circle CI test is failing is because this is what this PR is fixing! So, when this is merged then in the future that test will pass.

PR #319, opened 2020-01-20
feat: LT-1657 Display 'reject' instead of 'refuse' for open licences

stephanieboyd:


cases/forms/advice.py
```diff
@@ -1,7 +1,12 @@
 from lite_forms.components import Form, RadioButtons, Option, BackLink
 
 
-def advice_recommendation_form(post_url, back_url):
+def advice_recommendation_form(post_url, back_url, case_type):
+    if case_type == "open_licence":
```
Mark-Spiers-kainos: perhaps "application type? and use check if there is an application type enum?

templates/case/finalise-open-goods-countries.html
```diff
@@ -30,22 +31,38 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
...
 
 		<tbody class="govuk-table__body">
+            {% lcs 'cases.GoodsDecisionMatrixPage.NO_ADVICE_DEFAULT' as NO_ADVICE_DEFAULT%}
 			{% for good in case.application.goods_types %}
 				<tr class="govuk-table__row">
-					<td class="govuk-table__cell">{{ good.control_code|default:good.description }} <span style="opacity: 0.7">({{ good.advice.value|default:"No advice" }})</span></td>
+					<td class="govuk-table__cell">
+                        {{ good.control_code|default:good.description }}
+                        {% if country.advice.value == 'Refuse' and case.application.application_type.key == 'open_licence' %}
+                            <span style="opacity: 0.7">(Reject)</span>
```
Mark-Spiers-kainos: try not to use inline styles.

stephanieboyd: It was already there, I just had to change the text it was applied to. Do you have any suggestions on an alternative? :)

cases/forms/advice.py
stephanieboyd: changed, and there was an enum, thanks!

cases/forms/advice.py
```diff
@@ -1,23 +1,30 @@
...
-        "What do you advise?",
-        "You can advise to:",
+        cases.AdviceRecommendationForm.TITLE,
+        cases.AdviceRecommendationForm.DESCRIPTION,
         [
             RadioButtons(
                 "type",
                 [
                     Option("approve", "Grant the licence"),
                     Option("proviso", "Add a proviso"),
```
andytaylorkainos: Should these be in lite_content?

templates/case/finalise-open-goods-countries.html
andytaylorkainos: Should "(Reject)" be in lite_content?

stephanieboyd: Added

ui_automation_tests/conftest.py
```diff
@@ -212,7 +212,7 @@ def added_flags_on_queue(driver, context):  # noqa
     no = utils.get_element_index_by_text(elements, context.case_id, complete_match=False)
     driver.set_timeout_to(0)
     try:
-        if elements[no].find_element_by_css_selector(".lite-accordian-table__chevron").is_displayed():
+        if "3 of" in elements[no].text:
```
andytaylorkainos: hmmmm?

andytaylorkainos: Can this not be a bit better in terms of finding the correct <td>?

andytaylorkainos: Can this not be improved to find the correct `<td>` element?

andytaylorkainos: We could put an ID on the chevron element which contains the case-id (therefore it'd be unique and easily found)

stephanieboyd: @phlptt For your awareness

cases/forms/advice.py
stephanieboyd: Moved!

templates/case/finalise-open-goods-countries.html
stephanieboyd: Added

PR #332, opened 2020-01-27
feat: LT-1169 - MOD Clearance

jfaracik:
* Add case page for Exhibition Clearances
* Clean up a lot of status logic, moved old HTML page to LITE Forms view
* Added end to end test for exhibition clearances

RuopmXU


core/services.py
```diff
@@ -59,37 +58,34 @@ def get_countries(request, convert_to_options=False):
...
     user_permissible_statuses = user["user"]["role"]["statuses"]
+    statuses, _ = get_statuses(request)
 
     if case_type == CaseType.APPLICATION.value:
         case_type_applicable_statuses = [
             status
-            for status in get_statuses_as_options(request)
-            if status.value not in ["Applicant editing", "Closed", "Finalised", "Registered"]
+            for status in statuses
+            if status["key"] not in ["applicant_editing", "closed", "finalised", "registered"]
```
andytaylorkainos: can we have constants for those statuses please

core/services.py
```diff
@@ -59,37 +58,34 @@ def get_countries(request, convert_to_options=False):
...
         ]
     else:
         case_type_applicable_statuses = [
-            status
-            for status in get_statuses_as_options(request)
-            if status.value in ["Closed", "Submitted", "Withdrawn"]
+            status for status in statuses if status["key"] in ["closed", "submitted", "withdrawn"]
```
andytaylorkainos: can we have constants for those statuses please

templates/case/applications/exhibition-clearance.html
```diff
@@ -0,0 +1,223 @@
+{% extends 'layouts/case.html' %}
+
+{% load static humanize %}
+
+{% block details %}
+	<div class="app-case-board">
+		<div class="govuk-grid-row">
+			<div class="govuk-grid-column-one-quarter">
+				<p class="app-case-board__heading">{% lcs 'CASE_INFO_ORGANISATION' %}</p>
```
andytaylorkainos: Sadly you have to update the lite_content class structures for these strings (all of them in this file)

core/services.py
jfaracik: That really should have been raised in the pull request for that feature. I'll do it though.

templates/case/applications/exhibition-clearance.html
jfaracik: Seeing as none of the strings in the file are new and that this page is going to get nuked in a future story, could we just _not_?

andytaylorkainos: So long as you realise you're going to have to update the string classes within lite_content during that story?

jfaracik: Yeah of course. Every case page is going to be rewritten so it'd be the ideal time to do it

ui_automation_tests/step_defs/test_exhibition_clearance.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@when("I click on the Exhibition Clearance case")  # noqa
+def i_click_on_the_exhibition_clearance_case(driver, context):
+    driver.find_element_by_link_text(context.reference_code).click()
+
+
+@then("I see the Exhibition Clearance case page")  # noqa
+def i_see_the_exhibition_clearance_case_page(driver, context):
+    assert driver.find_element_by_id("heading-reference-code").text == context.reference_code
```
phlptt: could we store this selector in /pages/application_page especially as i imagine it will be used in other tests in the future now it's there.

PR #335, opened 2020-01-27
bug: LT-1898 Locations aren't showing when viewing a HMRC query

pkonopacki:
- included the section for locations to hmrc case page

- added an assertion to the UI test


templates/case/views/hmrc-goods-location.html
```diff
@@ -0,0 +1,17 @@
+<h3 class="govuk-heading-m">{% lcs 'CASE_GOODS_LOCATION' %}</h3>
+<table class="govuk-table" id="hmrc-goods-location">
+	<thead class="govuk-table__head">
+		<tr class="govuk-table__row">
+			<th scope="col" class="govuk-table__header govuk-!-width-one-third">{% lcs 'CASE_GOODS_LOCATION_NAME' %}</th>
```
andytaylorkainos: I know this will be confusing, but we have decided to follow a class structured approach when defining strings in lite_content.

Can you move the strings to the `HMRCPage.Good`class in `lite_content/lite_internal_frontend/cases.py`

ui_automation_tests/step_defs/test_hmrc_query.py
```diff
@@ -17,3 +20,4 @@ def go_to_hmrc(driver, context):
 def see_hmrc(driver, context):
     if context.hmrc_is_found:
         assert driver.find_element_by_id("case-type").text == "HMRC Query"
+        assert functions.element_with_id_exists(driver, ApplicationPage(driver).HMRC_GOODS_LOCATION)
```
andytaylorkainos: This only asserts that the table appears;
It'd be a better test to assert that the expected location appears within the table :)

templates/case/queries/hmrc-case.html
```diff
@@ -78,6 +78,9 @@ <h3 class="govuk-heading-m">{% lcs 'cases.HMRCPage.DenialReasons.TITLE' %}</h3>
 	    </table>
 	</form>
 
+	<!-- Goods Locations --> 
+	{% include 'case/views/hmrc-goods-location.html' %}
```
jfaracik: Seeing as this import is only used in `hmrc-case`, could you just inline it?

ui_automation_tests/step_defs/test_hmrc_query.py
jfaracik: I'm not too sure if the HMRC E2E is sophisticated enough to do that yet @andytaylorkainos - so I don't think it'll be in scope for this bug.

andytaylorkainos: The scope is to fix the bug. Tests validate that bugs are fixed. If @phlptt is happy with it as it is, then np!

andytaylorkainos: Nick seems happy enough :) ignore this request @pkonopacki

templates/case/views/hmrc-goods-location.html
pkonopacki: refactored strings for hmrc and standard applications, could you resolve and approve if you're happy @andytaylorkainos ?

templates/case/queries/hmrc-case.html
pkonopacki: done it, could you resolve and approve if you're happy @jfaracik ?

PR #336, opened 2020-01-28
feat: LT-974 - Not Yet Verified Good (System Flag)

andytaylorkainos:


templates/case/applications/open-licence-case.html
```diff
@@ -93,25 +94,16 @@ <h2 class="govuk-accordion__section-heading">
 									<td class="govuk-table__cell">{{ good.control_code|default_na }}</td>
 									<td class="govuk-table__cell">{{ good.is_good_incorporated|friendly_boolean }}</td>
 									<td class="govuk-table__cell">
+					                    {% for flag in good.flags %}
```
BenAndrew-Kainos: indentation looks a little messy here (maybe github but just double check)

andytaylorkainos: its github annoyingly

PR #338, opened 2020-01-29
feat: LT-1528 - respond to pv grading

Mark-Spiers-kainos:


cases/urls.py
```diff
@@ -42,8 +42,22 @@
     path("<uuid:pk>/finalise/", advice.Finalise.as_view(), name="finalise"),
     path("<uuid:pk>/ecju-queries/", ecju.ViewEcjuQueries.as_view(), name="ecju_queries"),
     path("<uuid:pk>/ecju-queries/add", ecju.CreateEcjuQuery.as_view(), name="ecju_queries_add"),
-    path("<uuid:pk>/respond-to-query/", goods_query.RespondCLCQuery.as_view(), name="respond_to_clc_query"),
-    path("<uuid:pk>/respond-to-query/flags/", goods_query.RespondFlags.as_view(), name="respond_to_clc_query_flags"),
+    path("<uuid:pk>/respond-to-clc-query/", goods_query.RespondCLCQuery.as_view(), name="respond_to_clc_query"),
+    path(
+        "<uuid:pk>/respond-to-clc-query/flags/",
+        goods_query.RespondCLCFlags.as_view(),
+        name="respond_to_clc_query_flags",
```
andytaylorkainos: Remove `_flags` please.
I know there are flags on the case which indicate that its a CLC query, but we don't respond to the flag (same for pv grading name too pls)

core/services.py
```diff
@@ -129,3 +130,16 @@ def get_control_list_entries(request, convert_to_options=False):
 
     data = get(request, CONTROL_LIST_ENTRIES_URL)
     return data.json()["control_list_entries"]
+
+
+def get_pv_gradings(request, convert_to_options=False):
+    data = get(request, GOV_PV_GRADINGS_URL)
+    if convert_to_options:
+
```
andytaylorkainos: remove blank line please

cases/views/goods_query.py
```diff
@@ -48,48 +44,105 @@ def post(self, request, **kwargs):
...
+    case = None
+    form = None
+
+    def dispatch(self, request, *args, **kwargs):
+        case_id = str(kwargs["pk"])
+        self.case = get_case(request, case_id)
+        self.form = respond_to_grading_query_form(request, self.case)
+
+        permissions = get_user_permissions(request)
+        if "RESPOND_PV_GRADING" not in permissions:
```
andytaylorkainos: constant please

cases/views/goods_query.py
```diff
@@ -98,7 +151,7 @@ def dispatch(self, request, *args, **kwargs):
         self.form = flags_form(flags=get_goods_flags(request, True), level="goods", origin="response", url="#")
         self.form.post_url = reverse_lazy("cases:respond_to_clc_query_flags", kwargs={"pk": self.case["id"]})
 
-        return super(RespondFlags, self).dispatch(request, *args, **kwargs)
+        return super(RespondCLCFlags, self).dispatch(request, *args, **kwargs)
```
andytaylorkainos: Remove flags from classname please (same for pv grading)

templates/case/queries/goods_query_case.html
```diff
@@ -3,120 +3,179 @@
...
-				<dt class="govuk-summary-list__key">
-                    {% lcs 'cases.ClcQuery.GoodDetails.FLAGS' %}
+            {% else %}
+                <dt class="govuk-summary-list__key">
+                    {% lcs 'cases.ClcQuery.GoodDetails.Clc.EXPECTED_CONTROL_CODE' %}
 				</dt>
+			{% endif %}
 				<dd class="govuk-summary-list__value">
-					{% include 'case/views/goods-flags.html' with origin='case' %}
+					{{ good.control_code|default:"Not on the control list" }}
```
andytaylorkainos: should the default be a lite_content string?

templates/case/queries/goods_query_case.html
```diff
@@ -3,120 +3,179 @@
...
 				</dd>
 			</div>
+			{% if good.is_good_controlled == "Unsure" %}
```
andytaylorkainos: Can you check that `good.is_good_controlled` isn't a key-value pair (it is in some areas of the application)

cases/urls.py
Mark-Spiers-kainos: This here is a page that is linked off the clc response overview page to set flags. so it does have to do with flags. if you have another name you would prefer happy to update.

cases/views/goods_query.py
Mark-Spiers-kainos: Same as url, This does have to do with the Flags, I can rename the class if you can think of a better name.

cases/urls.py
andytaylorkainos: makes sense now, name is good

cases/views/goods_query.py
andytaylorkainos: makes sense now, name is good

cases/views/goods_query.py
```diff
@@ -48,48 +46,105 @@ def post(self, request, **kwargs):
             form.post_url = reverse_lazy("cases:respond_to_clc_query_flags", kwargs={"pk": self.case["id"]})
             return form_page(request, form, data={"flags": self.case["query"]["good"]["flags"]})
 
-        if request.POST.get("action") != "change":
-            form_data = request.POST.copy()
+        if request.POST.get("action") == "change":
```
andytaylorkainos: are these actions constant throughout the system or do they change per view/form?

cases/views/goods_query.py
```diff
@@ -48,48 +46,105 @@ def post(self, request, **kwargs):
...
+        )
 
+        if response:
+            return response
+
+        if not request.POST.get("validate_only"):
             return redirect(reverse_lazy("cases:case", kwargs={"pk": self.case["id"]}))
 
+        response_data = response_data["control_list_classification_query"]
+        # If validate only is removed (therefore the user is on the overview page
```
andytaylorkainos: should this message be before line 64?

cases/views/goods_query.py
```diff
@@ -48,48 +46,105 @@ def post(self, request, **kwargs):
...
+
+    def get(self, request, **kwargs):
+        return form_page(request, self.form)
+
+    def post(self, request, **kwargs):
+        # If 'set-flags' take them to the goods flags page
+        if request.POST.get("action") == "set-flags":
+            form = flags_form(flags=get_goods_flags(request, True), level="goods", origin="response", url="#")
+
+            form.questions.append(HiddenField("prefix", request.POST["prefix"]))
```
andytaylorkainos: picky, but can we move this logic to a separate method to tidy the view up a little?

cases/views/goods_query.py
```diff
@@ -48,48 +46,105 @@ def post(self, request, **kwargs):
...
+            put_goods_query_pv_grading,
+            object_pk=str(self.case["query"]["id"]),
+            override_data=form_data,
+        )
+
+        if response:
+            return response
+
+        if not request.POST.get("validate_only"):
+            return redirect(reverse_lazy("cases:case", kwargs={"pk": self.case["id"]}))
```
andytaylorkainos:
``` 
       if response:
            return response

        if not request.POST.get("validate_only"):
            return redirect(reverse_lazy("cases:case", kwargs={"pk": self.case["id"]}))
```

This seems like a pattern that you intend to repeat, can you create a helper library for it?

cases/views/goods_query.py
```diff
@@ -112,9 +167,37 @@ def post(self, request, **kwargs):
...
+        return render(request, "case/queries/clc_query_response_overview.html", context)
+
+
+class RespondPVGradingFlags(TemplateView):
+    case = None
+    form = None
+
+    def dispatch(self, request, *args, **kwargs):
+        self.case = get_case(request, str(kwargs["pk"]))
+        self.form = flags_form(flags=get_goods_flags(request, True), level="goods", origin="response", url="#")
```
andytaylorkainos: super picky, so only asking you to do it for code you've just written -
Can you create an enum for the flag levels please?

core/services.py
```diff
@@ -129,3 +130,15 @@ def get_control_list_entries(request, convert_to_options=False):
...
     return data.json()["control_list_entries"]
+
+
+def get_pv_gradings(request, convert_to_options=False):
+    data = get(request, GOV_PV_GRADINGS_URL)
+    if convert_to_options:
+        converted_units = []
+        for pvg in data.json().get("pv_gradings"):
+            for key in pvg:
+                converted_units.append(Option(key=key, value=pvg[key],))
```
andytaylorkainos: remove that stray `,` please

cases/views/goods_query.py
Mark-Spiers-kainos: They only exist in the context of these respond to query form. If these types of forms become more common we might make a new form class/formview class.

cases/views/goods_query.py
Mark-Spiers-kainos: It is only used for these forms since we had to overwrite the default post action. Otherwise this code is in there.

cases/forms/respond_to_good_query.py
```diff
@@ -1,54 +1,101 @@
...
+    )
+
+
+def respond_to_grading_query_form(request, case):
+    return Form(
+        title=cases.RespondGradingQueryForm.TITLE,
+        questions=[
+            Heading(case["id"], HeadingStyle.S),
+            SummaryList(values={"Description": case["query"]["good"]["description"]}),
+            HTMLBlock(html='<hr class="lite-horizontal-separator">'),
```
BenAndrew-Kainos: I know it's nit-picky but could you create a horizontal separator component instead

cases/forms/respond_to_good_query.py
```diff
@@ -1,54 +1,101 @@
...
+            Group(
+                name="grading",
+                components=[
+                    TextInput(title=cases.RespondGradingQueryForm.Grading.PREFIX, name="prefix", optional=True),
+                    Select(
+                        options=get_pv_gradings(request=None, convert_to_options=True),
```
BenAndrew-Kainos: Why is request none?

cases/views/goods_query.py
```diff
@@ -38,67 +36,130 @@ def get(self, request, **kwargs):
...
         if response_data.get("report_summary"):
             context["report_summary"] = get_picklist_item(request, response_data["report_summary"])
 
-        return render(request, "case/clc_query_overview.html", context)
+        return render(request, "case/queries/clc_query_response_overview.html", context)
+
+    def display_flag_form(self, request):
+        form = flags_form(flags=get_goods_flags(request, True), level=FlagLevels.GOODS, origin="response", url="#")
+
+        form.questions.append(HiddenField("is_good_controlled", request.POST["is_good_controlled"]))
```
BenAndrew-Kainos: for loop for these please

cases/views/goods_query.py
```diff
@@ -38,67 +36,130 @@ def get(self, request, **kwargs):
...
         }
 
         if response_data.get("report_summary"):
             context["report_summary"] = get_picklist_item(request, response_data["report_summary"])
 
-        return render(request, "case/clc_query_overview.html", context)
+        return render(request, "case/queries/clc_query_response_overview.html", context)
+
+    def display_flag_form(self, request):
+        form = flags_form(flags=get_goods_flags(request, True), level=FlagLevels.GOODS, origin="response", url="#")
```
BenAndrew-Kainos: Why is the url # ?

cases/views/goods_query.py
```diff
@@ -38,67 +36,130 @@ def get(self, request, **kwargs):
...
+        response_data = response_data["pv_grading_query"]
+
+        context = {
+            "data": response_data,
+            "case": self.case,
+        }
+
+        return render(request, "case/queries/pv_grading_query_response_overview.html", context)
+
+    def display_flag_form(self, request):
```
BenAndrew-Kainos: Same/Similar code again. Could the be refactored to share one function

templates/case/queries/pv_grading_query_response_overview.html
```diff
@@ -0,0 +1,119 @@
+{% extends 'layouts/two-pane.html' %}
+
+{% block back_link %}{% endblock %}
+
+{% block two_thirds %}
+
+	<h2 class="govuk-heading-l">{% lcs 'cases.GradingResponseOverview.TITLE' %}</h2>
+
+    <p class="govuk-body-m">{% lcs 'cases.GradingResponseOverview.WARNING' %}</p>
```
BenAndrew-Kainos: Indentation here seems inconsistent

templates/case/queries/pv_grading_query_response_overview.html
```diff
@@ -0,0 +1,119 @@
+{% extends 'layouts/two-pane.html' %}
+
+{% block back_link %}{% endblock %}
+
+{% block two_thirds %}
+
+	<h2 class="govuk-heading-l">{% lcs 'cases.GradingResponseOverview.TITLE' %}</h2>
```
BenAndrew-Kainos: Accessibility requires aa H1

cases/forms/respond_to_good_query.py
Mark-Spiers-kainos: Because we don't need the user token to get data from static endpoints.

cases/views/goods_query.py
Mark-Spiers-kainos: There isn't a nice way to return to the original form, with all the data intact, if you can think of one I will implement it.

cases/views/goods_query.py
Mark-Spiers-kainos: This code only exists in the premise of this overview page, and the flags are limited to goods_query overview page. I don't expect this code to be replaced anywhere else so would prefer to leave it in the classes, easy to manage.

cases/views/goods_query.py
Mark-Spiers-kainos: The url in this case is referring to the backlink.

cases/forms/respond_to_good_query.py
BenAndrew-Kainos: Could you leave a comment

core/builtins/custom_tags.py
```diff
@@ -277,3 +277,12 @@ def username(user: dict):
         return user["first_name"] + " " + user["last_name"]
 
     return user["email"]
+
+
+@register.filter()
+def display_grading(text: str):
```
andytaylorkainos: why do we need this filter? should the permission not be returned as {"key: "value"}?

Mark-Spiers-kainos: The form holds the data from the post request, which only contains the key.

PR #340, opened 2020-01-31
feat: LT-1084 - Record details of when a customs enquiry is being performed

jfaracik:
* Add have_goods_departed field to HMRC case
* Remove control list entry related thing from HMRC case
* Updated to the latest shared assets and forms


assets/css/overrides/_table.scss
```diff
@@ -8,3 +8,8 @@
 		}
 	}
 }
+
+.govuk-table__caption {
+	@include govuk-font($size: 27, $weight: bold);
+	margin-bottom: govuk-spacing(3);
+}
```
BenAndrew-Kainos: newline

templates/case/queries/hmrc-case.html
```diff
@@ -38,33 +37,30 @@ <h3 class="govuk-heading-m">{% lcs 'cases.HMRCPage.DenialReasons.TITLE' %}</h3>
 			{% lcs 'cases.HMRCPage.Good.EDIT_FLAGS' %}
 		</button>
 	    <table class="govuk-table">
+			<caption class="govuk-table__caption">{% lcs 'cases.HMRCPage.Good.TITLE' %}</caption>
 	        <thead class="govuk-table__header">
-	            <tr>
+	            <tr class="govuk-table__row">
+					<th class="govuk-table__header" scope="col">
+						<button type="button" name="button" class="lite-button-checkbox" title="Select all/Deselect all"></button>
```
BenAndrew-Kainos: Can we move the "select all/deselect all" to a generic string

BenAndrew-Kainos:
Is it worth adding to an e2e to verify they can se that new HTML?

jfaracik:
> Is it worth adding to an e2e to verify they can se that new HTML?

It really depends on how in-depth we want E2Es to be. We haven't tested every combination elsewhere, so it'd be unusual to start now.

BenAndrew-Kainos:
> > Is it worth adding to an e2e to verify they can se that new HTML?
> 
> It really depends on how in-depth we want E2Es to be. We haven't tested every combination elsewhere, so it'd be unusual to start now.

I'd disagree that we haven't done this before but you are right in that we're inconsistent. Until anyone says otherwise we'll leave it

PR #347, opened 2020-02-05
chore: Set dependency versions

phlptt:
* Set dependency versions based on what was in the current master Pipfile.lock
* Moved some dependencies from packages to dev packages
* got rid of a couple of packages that are no longer used.


phlptt:
https://jenkins.ci.uktrade.io/job/lite-e2e-internal-frontend/976/console
tests running.

PR #350, opened 2020-02-05
LT-1658: Expose inactive parties on case page

a-bek:
- Inactive parties table in case detail
- e2e tests


templates/case/includes/inactive-parties.html
```diff
@@ -0,0 +1,42 @@
...
+						{% for flag in party.flags %}
+							{{ flag.name }}<br>
+						{% endfor %}
+					</td>
+				</tr>
+			{% endfor %}
+		</tbody>
+	</table>
+{% else %}
+	<p class="govuk-hint">
```
jfaracik: Could you use the notice control?

Works like so:

`{% include "includes/notice.html" with text='CASE_ENTITIES_NONE' %}

ui_automation_tests/pages/application_page.py
```diff
@@ -35,6 +35,7 @@ class ApplicationPage(BasePage):
     AUDIT_TRAIL_ITEM = ".app-activity__item"  # CSS
     APPLICATION_SUMMARY_BOARD = ".app-case-board"  # CSS
     TABLE_ENTITIES = "table-entities"  # ID
+    TABLE_INACTIVE_ENTITIES = "table-inactive-entities"  # ID
```
jfaracik: Suffix the constant with `_ID` and remove the comment :)

templates/case/applications/standard-licence-case.html
```diff
@@ -186,6 +186,20 @@ <h2 class="govuk-accordion__section-heading">
 			</div>
 		</div>
 
+		<!-- Inactive Parties -->
+		<div class="govuk-accordion__section">
+			<div class="govuk-accordion__section-header">
+				<h2 class="govuk-accordion__section-heading">
+					<span class="govuk-accordion__section-button" id="accordion-default-heading-3">
+						{% lcs 'CASE_ENTITIES_INACTIVE' %}
```
andytaylorkainos: you'll need to change this string to use the class approach (`CASES.something.ENTITIES_INACTIVE`)

templates/case/includes/inactive-parties.html
```diff
@@ -0,0 +1,42 @@
...
+		{% lcs 'CASE_ENTITIES_NONE' %}
```
andytaylorkainos: same again for the class approach

templates/case/includes/inactive-parties.html
```diff
@@ -0,0 +1,42 @@
...
+	</p>
+{% endif %}
```
andytaylorkainos: newline

PR #355, opened 2020-02-10
Bug: LT-2066 - editing admin team

Mark-Spiers-kainos:


teams/constants.py
```diff
@@ -0,0 +1,5 @@
+from enum import Enum
+
+
+class SystemTeamsID(Enum):
+    ADMIN = "00000000-0000-0000-0000-000000000001"
```
efahy: Up to now all constants have gone into `conf/constants.py` and I think that's fine.  Can you put this in there, rather than creating a whole new module that will probably only ever contain one or two constants?

PR #361, opened 2020-02-12
feat: LT-972 - copy application

Mark-Spiers-kainos:


templates/case/includes/parties.html
```diff
@@ -113,9 +113,11 @@
                 <td class="govuk-table__cell">{{ party.sub_type.value }} ({% lcs 'cases.ApplicationPage.ThirdParty.ROLE' %}{{ party.role.value  }})</td>
                 <td class="govuk-table__cell">{{ party.website|linkify }}</td>
                 <td class="govuk-table__cell">
-                    <a href="{% url 'cases:document' case.id party.document.id %}">
-                        {{ party.document.name }}
-                    </a>
+					{% if party.document %}
```
Mark-Spiers-kainos: was throwing error for third parties since they don't require a document

templates/case/queries/end_user_advisory.html
```diff
@@ -119,16 +119,6 @@
 					{{ case.query.note|default_na }}
 				</dd>
 			</div>
-            {% if case.query.copy_of %}
```
Mark-Spiers-kainos: this is moved to top of case page. Same place for every case type

ui_automation_tests/features/copied_application.feature
```diff
@@ -0,0 +1,9 @@
+@all @internal @copied_applications
+Feature: I want to see that a copied application references the previous application
+
+  @LT_972_copy_application @regression
+  Scenario: View original application link in new application
+    Given I sign in to SSO or am signed into SSO
+    And I have an open application from copying
+    When I go to application previously created
+    Then I can see the original application is linked
```
BenAndrew-Kainos: newline

BenAndrew-Kainos: Also you may want to check some properties are copied like name etc.

ui_automation_tests/step_defs/test_copy_application.py
```diff
@@ -0,0 +1,10 @@
+from pytest_bdd import then, scenarios
+
+from ui_automation_tests.pages.application_page import ApplicationPage
+
+scenarios("../features/copied_application.feature", strict_gherkin=False)
+
+
+@then("I can see the original application is linked")
+def original_application_linked(driver, context):
+    assert context.old_app_id in ApplicationPage(driver).get_case_copy_of_field().get_attribute("href")
```
BenAndrew-Kainos: I'd move the `get_attribute("href")` bit to the page

ui_automation_tests/features/copied_application.feature
Mark-Spiers-kainos: This test was just developed due to the one element being tested in it, other than that we have no reason for the test, cause we expect the API to test it thoroughly.

PR #362, opened 2020-02-13
Lt 1177 users pagination

a-bek:
- Allow gov users to be filtered by status
- Return gov_user filters from cases api rather than make an additional request to API


ui_automation_tests/step_defs/test_zz_users.py
```diff
@@ -36,7 +36,7 @@ def test_manage_users(driver, open_internal_hub, context, internal_info):
     user_page.click_save_and_continue()
 
     assert (
-        driver.find_element_by_tag_name("h2").text == "Users"
+        driver.find_element_by_tag_name("h1").text == "Users"
```
Mark-Spiers-kainos: If this is checking against lite_content strings, can you remove it? And perhaps test it in some other way if not already covered enough by the test

cases/forms/assign_users.py
```diff
@@ -6,10 +6,11 @@
 
 
 def assign_users_form(request: HttpRequest, team_id, queue, multiple: bool):
+    params = {"teams": team_id, "no_page": True}
```
BenAndrew-Kainos: does "no_page" mean page number (if so could you call it page like the other paginated pages)

templates/users/change_status.html
```diff
@@ -7,9 +7,9 @@ <h2 class="govuk-heading-l">{{ title }}</h2>
 		<div class="buttons-row">
 			{% csrf_token %}
 			{% if status == 'reactivate' %}
-				<button type="submit" name="status" value="Active" class="govuk-button">Reactivate User</button>
+				<button type="submit" name="status" value="Active" class="govuk-button" id="reactivated_button">Reactivate User</button>
```
BenAndrew-Kainos: Move strings to content (same for rest of file)

templates/users/index.html
```diff
@@ -3,24 +3,25 @@
 {% block back_link %}{% endblock %}
 
 {% block body %}
-
 	<div class="lite-app-bar">
 		<div class="lite-app-bar__content">
-			<h2 class="govuk-heading-l">{{ title }}</h2>
+			<h1 class="govuk-heading-l">{{ title }}</h1>
```
BenAndrew-Kainos: Nice accessibility change :)

ui_automation_tests/pages/users_page.py
```diff
@@ -14,6 +17,10 @@ class UsersPage(BasePage):
     TEAM = "team"
     ROLE = "role"
     EDIT_BUTTONS_IN_TABLE = '.govuk-table__cell a[href*="edit"]'
+    DEACTIVATE_BUTTON = '.govuk-button[href*="deactivate"]'  # CSS
```
BenAndrew-Kainos: Jan says we don't do these comments anymore and instead we include CSS_SELECTOR in the variable name 😢

BenAndrew-Kainos: Also could you give all buttons an ID please

ui_automation_tests/step_defs/test_zz_users.py
BenAndrew-Kainos: Yeah we shouldn't ever assert content in e2e's (although I understand it has been done in the past)

users/views/users.py
```diff
@@ -18,11 +20,32 @@
...
+        status = request.GET.get("status", "active")
+        params = {"page": int(request.GET.get("page", 1)), "status": status}
+
+        data, _ = get_gov_users(request, params)
+
         user, _ = get_gov_user(request, str(request.user.lite_api_user_id))
         super_user = is_super_user(user)
 
-        context = {"data": data, "title": "Users", "super_user": super_user}
+        # Return from API and swap key and value names.
```
BenAndrew-Kainos: Is this a TODO?

users/views/users.py
```diff
@@ -18,11 +20,32 @@
...
+        statuses = [
+            Option(option["key"], option["value"])
+            for option in [{"key": "active", "value": "Active"}, {"key": "", "value": "All"}]
```
BenAndrew-Kainos: Move these to constant list or static API request (it sounds like your planning to do that anyway from the comment above)

cases/forms/assign_users.py
iichr: It means no pagination needed :)

users/views/users.py
a-bek: Moved filters to API.

templates/users/index.html
```diff
@@ -42,11 +43,11 @@ <h2 class="govuk-heading-l">{{ title }}</h2>
 							</a>
 						{% else %}
 							<strong class="govuk-tag govuk-phase-banner__content__tag ">
-								Pending
+								{% lcs 'Users.PENDING' %}
 							</strong>
 						{% endif %}
                     </td>
```
jfaracik: `{{ user.email }}` should be in quotes :)

a-bek: Eyes of a hawk

PR #367, opened 2020-02-14
Seed hmrc 2

phlptt:


ui_automation_tests/conftest.py
```diff
@@ -125,7 +125,7 @@ def click_on_created_eua(driver, context, internal_url):  # noqa
     driver.get(internal_url.rstrip("/") + "/cases/" + context.eua_id)
 
 
-@given("I create application or application has been previously created")  # noqa
+@given("I create standard application or standard application has been previously created")  # noqa
```
BenAndrew-Kainos: grammar "I create standard application or **a** standard application has been previously created"

ui_automation_tests/step_defs/test_hmrc_query.py
```diff
@@ -1,23 +1,22 @@
...
-        context.hmrc_is_found = True
-        driver.find_element_by_css_selector('.govuk-table__body .govuk-link[href*="cases"]').click()
-    driver.set_timeout_to(10)
+@given("I create HMRC query")
+def create_hmrc(apply_for_hmrc_query):
+    pass
+
+
+@when("I go to HMRC query previously created")
+def go_to_hmrc(driver, internal_url, context):
```
BenAndrew-Kainos: Theres already a "I go to case" when in conftest which is identical so just use that

ui_automation_tests/step_defs/test_hmrc_query.py
```diff
@@ -1,23 +1,22 @@
...
+    driver.get(internal_url.rstrip("/") + "/cases/" + context.case_id)
 
 
 @then("I see HMRC query")
 def see_hmrc(driver, context):
-    if context.hmrc_is_found:
-        assert driver.find_element_by_id("case-type").text == "HMRC Query"
-        assert functions.element_with_id_exists(driver, ApplicationPage(driver).HMRC_GOODS_LOCATION)
+    assert driver.find_element_by_id("case-type").text == "HMRC Query"
```
BenAndrew-Kainos: This should be in a page

PR #370, opened 2020-02-21
feat: LT-2149 - Add PV Grading and CLC to list of available goods query states

iichr:
- added two additional states for goods queries - PV review and CLC review

https://uktrade.atlassian.net/browse/LT-2149

### DIFF
--- a/core/services.py
+++ b/core/services.py
@@ -11,6 +11,7 @@ from conf.constants import (
     STATUS_PROPERTIES_URL,
     GOV_PV_GRADINGS_URL,
     Statuses,
+    GOODS_QUERY_STATUSES,
 )
 from users.services import get_gov_user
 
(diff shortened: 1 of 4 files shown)


core/services.py
```diff
@@ -88,7 +88,9 @@ def get_permissible_statuses(request, case_type):
         ]
     else:
         case_type_applicable_statuses = [
-            status for status in statuses if status["key"] in [Statuses.CLOSED, Statuses.SUBMITTED, Statuses.WITHDRAWN]
+            status
+            for status in statuses
+            if status["key"] in [Statuses.CLOSED, Statuses.SUBMITTED, Statuses.WITHDRAWN, Statuses.CLC, Statuses.PV]
```
BenAndrew-Kainos: Could you define this list as a constant (perhaps in constants.py)

PR #371, opened 2020-02-21
test: Tidy up flags test

phlptt:
* Seed the flag via API so that it is more reliable and quicker
* Update to latest forms and content submodules
* fix because of new content
* Tidy up of users tests

https://jenkins.ci.uktrade.io/job/lite-e2e-internal-frontend/1125/

blue build here


ui_automation_tests/step_defs/test_assign_case_flags_to_case.py
```diff
@@ -14,6 +14,12 @@ def click_edit_case_flags_link(driver):
     application_page.click_edit_case_flags()
 
 
+@when("I click edit destination flags link")
+def click_edit_case_flags_link(driver):
+    application_page = ApplicationPage(driver)
```
BenAndrew-Kainos: This could be oneline

ui_automation_tests/step_defs/test_users.py
```diff
@@ -0,0 +1,63 @@
...
+
+@then("I see new user")
+def see_new_user(driver, context):
+    assert utils.paginated_item_exists(context.added_email, driver), "Item couldn't be found"
+
+
+@when("I deactivate new user")
+def deactivate_user(driver, context):
+    user_page = UsersPage(driver)
+    user_page.go_to_users_page(context)
```
BenAndrew-Kainos: Why do you use user_page to go to user page (seems odd)

ui_automation_tests/step_defs/test_assign_case_flags_to_case.py
phlptt: done :)

ui_automation_tests/step_defs/test_users.py
phlptt: all of this has just been moved but i'll change it

phlptt: actually it does make sense ists users page so the users list

PR #378, opened 2020-02-25
feat: LT-2032 Exhibition Details

rorysproule1:


cases/services.py
```diff
@@ -354,7 +354,10 @@ def _generate_post_data_and_errors(keys, request_data, action):
 def _get_total_goods_value(case):
     total_value = 0
     for good in case.get("application").get("goods", []):
-        total_value += Decimal(good["value"]).quantize(Decimal(".01"))
+        if good.get("value"):
+            total_value += Decimal(good["value"]).quantize(Decimal(".01"))
+        else:
+            return None
```
BenAndrew-Kainos: What about if you have a mixture of goods with values & without values (i.e. some software & some physical). I'd suggest removing the else return None bit

core/builtins/custom_tags.py
```diff
@@ -67,6 +67,13 @@ def str_date(value):
     )
 
 
+@register.filter
+@stringfilter
+def str_date_only(value):
+    return_value = do_timezone(datetime.datetime.strptime(value, DATE_NO_TIME), "Europe/London")
+    return return_value.strftime("%d %B " "%Y")
```
BenAndrew-Kainos: What's up with this string here (why is it two strings?)

rorysproule1: Corrected to one string.

cases/services.py
rorysproule1: Removed the else.

cases/services.py
```diff
@@ -354,7 +354,8 @@ def _generate_post_data_and_errors(keys, request_data, action):
 def _get_total_goods_value(case):
     total_value = 0
     for good in case.get("application").get("goods", []):
-        total_value += Decimal(good["value"]).quantize(Decimal(".01"))
+        if good.get("value"):
```
stephanieboyd: Comment as to why it can be zero would be good here

conf/constants.py
```diff
@@ -1,6 +1,7 @@
 from enum import Enum
 
 ISO8601_FMT = "%Y-%m-%dT%H:%M:%S.%fZ"
+DATE_NO_TIME = "%Y-%m-%d"
```
stephanieboyd: minor - `DATE_FORMAT` might have been a better name. it's not a date, it's a format :)

core/builtins/custom_tags.py
```diff
@@ -67,6 +67,13 @@ def str_date(value):
     )
 
 
+@register.filter
+@stringfilter
+def str_date_only(value):
+    return_value = do_timezone(datetime.datetime.strptime(value, DATE_NO_TIME), "Europe/London")
```
stephanieboyd: minor - return_value isn't a good name, call it what it is please :)

templates/case/applications/mod-clearance.html
```diff
@@ -57,6 +49,39 @@
 	</div>
 
 	<div class="govuk-accordion app-accordion" data-module="govuk-accordion" id="accordion-default">
+		{% if case.case_type.sub_type.key == "exhibition_clearance"  %}
```
stephanieboyd: Is there not a constant for these sub types? If there isn't, there probably should be as it's used a lot and in different files

cases/services.py
rorysproule1: Added a comment to explain.

conf/constants.py
rorysproule1: Changed to DATE_FORMAT

core/builtins/custom_tags.py
rorysproule1: Changed to date_str

templates/case/applications/mod-clearance.html
rorysproule1: Added a constant

PR #387, opened 2020-03-02
Feat: Lt 980 edit company name for organisation

stephanieboyd:
- Add edit organisation


organisations/views.py
```diff
@@ -125,3 +138,27 @@ def post(self, request, **kwargs):
 
         messages.success(request, strings.HMRC_ORGANISATION_CREATION_SUCCESS)
         return redirect("organisations:hmrc")
+
+
+class EditBusiness(MultiFormView):
+    forms = None
+
+    def __init__(self, **kwargs):
```
Mark-Spiers-kainos: I don't like having two inits, especially when it can all be moved into one.

organisations/forms.py
```diff
@@ -119,3 +120,76 @@ def register_hmrc_organisation_forms(name=""):
...
+                            Option(key="commercial", value="Commercial"),
+                            Option(key="individual", value="Individual"),
+                        ],
+                    )
+                ],
+                default_button_name=strings.CONTINUE,
+            ),
+            conditional(
+                individual,
+                Form(
```
Mark-Spiers-kainos: perhaps move the forms inside the conditional to separate functions, to be a bit easier to read.

organisations/forms.py
```diff
@@ -119,3 +120,76 @@ def register_hmrc_organisation_forms(name=""):
...
+                    title=strings.RegisterBusiness.EDIT_COMMERCIAL_TITLE,
+                    questions=[
+                        conditional(
+                            permission_to_edit_org_name,
+                            TextInput(title=strings.REGISTER_BUSINESS_FIRST_AND_LAST_NAME, name="name"),
+                        ),
+                        TextInput(title=strings.RegisterBusiness.EORI_NUMBER, name="eori_number"),
+                        TextInput(
+                            title=strings.RegisterBusiness.SIC_NUMBER,
+                            description="Classifies industries by a four-digit code.",
```
andytaylorkainos: What is dis string

stephanieboyd: Fixed. It was actually duplicated, so I added the correct string

PR #390, opened 2020-03-03
feat: LT-1385 - Highlight cases that haven't been updated in the last five days

jfaracik:
* Added indicator on cases screen for when a case hasn't been updated in at least five days
* Added tooltip to explain indicator
* Added styling


cases/views/main.py
```diff
@@ -106,6 +106,7 @@ def get(self, request, **kwargs):
             "params_str": convert_dict_to_query_params(params),
             "updated_cases_banner_queue_id": updated_cases_banner_queue_id,
             "filters": filters,
+            "is_all_cases_queue": queue_id == DEFAULT_QUEUE_ID,
```
a-bek: 👏

PR #398, opened 2020-03-06
fix: LT-2131 - display document name instead of download on case page

cstobart:


templates/case/applications/standard-licence-case.html
```diff
@@ -108,7 +108,7 @@ <h2 class="govuk-accordion__section-heading">
 					                        {% for document in good_on_application.good.documents %}
 					                            {% if document.safe == True %}
 					                                <a class="govuk-link govuk-link--no-visited-state" id="good_document" href="{% url 'cases:document' case.id document.id %}">
-					                                    {% lcs 'cases.Manage.Documents.DOWNLOAD_DOCUMENT' %}
+					                                    {{ document.name }}
```
jfaracik: Prefix the file name with "Download"

templates/case/views/standard-goods.html
```diff
@@ -40,7 +40,7 @@ <h3 class="govuk-heading-m">{% lcs 'cases.ApplicationPage.Goods.TITLE' %}</h3>
                         {% for document in good_on_application.good.documents %}
                             {% if document.safe == True %}
                                 <a class="govuk-link govuk-link--no-visited-state" id="good_document" href="{% url 'cases:document' case.id document.id %}">
-                                    {% lcs 'cases.Manage.Documents.DOWNLOAD_DOCUMENT' %}
+                                    {{ document.name }}
```
jfaracik: Prefix the file name with "Download"

PR #399, opened 2020-03-09
feat: LT-1024 - Register an organisation

jfaracik:
* Organisations page - added `Status` column
* E2E fixes
* EXTRA: Removed some styling from `placeholder.scss`
* EXTRA: Fixed Teams page header


jfaracik:
> Change the name of the PR, it doesn't correspond to what it actually does

Our PRs have always followed this format

PR #404, opened 2020-03-11
Lt 1295 assign user and work queue

stephanieboyd:


ui_automation_tests/conftest.py
```diff
@@ -333,3 +333,9 @@ def go_to_users(driver, sso_sign_in, internal_url):  # noqa
 @given("an Exhibition Clearance is created")  # noqa
 def an_exhibition_clearance_is_created(driver, apply_for_exhibition_clearance):
     pass
+
+
+@when("I click assign user Button")
```
phlptt: if this is only use by one step_def file, it doesn't need to be in conftest

stephanieboyd: changed

cases/urls.py
```diff
@@ -92,4 +92,6 @@
         name="assign_destination_flags",
     ),
     path("<uuid:pk>/case-officer/", main.CaseOfficer.as_view(), name="case_officer"),
+    path("<uuid:pk>/assign-user/", main.UserWorkQueue.as_view(), name="assign_user"),
```
andytaylorkainos: both these endpoints assign users to a queue. Would it not be better to have a more generic endpoint with a query parameter to figure out which logic to execute?

cases/views/main.py
```diff
@@ -360,3 +361,29 @@ def post(self, request, **kwargs):
...
+
+
+class UserWorkQueue(SingleFormView):
+    def init(self, request, **kwargs):
+        self.object_pk = kwargs["pk"]
+        self.form = assign_user_and_work_queue(request)
+        self.action = get_gov_user_from_form_selection
+
+    @staticmethod
+    def _get_form_data(request, case_pk, json):
```
andytaylorkainos: are you sure you have to do this hack; have you spoken to Jan?

queues/services.py
```diff
@@ -61,6 +63,16 @@ def get_queue_case_assignments(request, pk):
     return data.json(), data.status_code
 
 
-def put_queue_case_assignments(request, pk, json):
-    data = put(request, QUEUES_URL + pk + "/case-assignments/", json)
-    return data.json(), data.status_code
+def put_queue_case_assignments(request, pk, json, single_case=True):
+    # single_case being true is a flag that indicates an assignment is being set from within the case actions
+    if single_case:
```
andytaylorkainos: do you need this flag, if you can just use the code below? `queue = json.get("flag")`

cases/views/main.py
stephanieboyd: I spoke with both Ben and Mark, and found the example of this in our code already. Seems to be the way we approach it.

cases/urls.py
stephanieboyd: Decent suggestion, pushing back for now - it would require lifting and shifting lots of logic, and changing test code.

queues/services.py
stephanieboyd: I had a think about your comment. If I don't have this flag, then it's check if there is a queue in the json and if not, check if there is a pk and then error out. To me, this flow is a little more readable.

andytaylorkainos: Can you make it a separate function instead then please, instead of going into a function with a pre-set flag which branches into separate logic

PR #410, opened 2020-03-17
feat: LT-1264 Completed Queue's Workflow

BenAndrew-Kainos:


ui_automation_tests/step_defs/test_assign_users_to_queue.py
```diff
@@ -145,3 +147,36 @@ def case_officer_is_set(driver, internal_info):
...
+    ApplicationPage(driver).click_im_done_button()
+
+
+@when("I unassign myself for my work queue")
+def unassign_queue(driver, context):
+    UnassignQueuePage(driver).check_unassign_checkbox(context.queue_name)
+    functions.click_submit(driver)
+
+
+@when("I go to my work queue")
```
phlptt: might be worth renaming this to my newly created work queue, as my work queue is a specific system queue also

BenAndrew-Kainos: Oh right will do

cases/urls.py
```diff
@@ -7,6 +7,8 @@
 urlpatterns = [
     path("", main.Cases.as_view(), name="cases"),
     path("<uuid:pk>/", main.ViewCase.as_view(), name="case"),
+    path("<uuid:pk>/done/", main.CaseProcessedByUser.as_view(), name="done"),
+    path("<uuid:pk>/done/<uuid:queue_id>", main.CaseProcessedByUserForQueue.as_view(), name="done_for_queue"),
```
andytaylorkainos: i'd prefer it to be a query parameter :( - unless the Views inherit from different classes

cases/views/main.py
```diff
@@ -205,6 +225,29 @@ def post(self, request, **kwargs):
...
+        else:
+            return reverse_lazy("cases:cases")
+
+
+class CaseProcessedByUserForQueue(TemplateView):
+    def get(self, request, pk, queue_id):
+        _, status_code = put_unassign_queues(request, str(pk), {"queues": [str(queue_id)]})
+        if status_code != HTTPStatus.OK:
+            # If this fails, revert to standard unassign view
+            return redirect(reverse_lazy("cases:done", kwargs={"pk": pk}))
```
andytaylorkainos: we don't want to display the reason it failed?

cases/urls.py
BenAndrew-Kainos: they inherit from different classes

cases/views/main.py
BenAndrew-Kainos: Changed it to redirect to the case page but I can't show errors

cases/services.py
```diff
@@ -54,6 +54,17 @@ def put_case(request, pk, json):
     return data.json(), data.status_code
 
 
+# Queue assignment actions
+def get_user_case_queues(request, pk):
+    data = get(request, CASE_URL + str(pk) + "/assigned-queues/")
+    return data.json(), data.status_code
+
+
+def put_unassign_queues(request, pk, json):
```
stephanieboyd: Unassign?

cases/views/main.py
```diff
@@ -142,11 +145,25 @@ def get(self, request, **kwargs):
...
+        is_system_queue = True
+
+        if queue_id:
+            queue = get_queue(request, queue_id)
+            if queue.get("queue"):
+                is_system_queue = queue["queue"].get("is_system_queue", True)
 
         if "application" in case:
             status_props, _ = get_status_properties(request, case["application"]["status"]["key"])
+            can_set_done = (
```
stephanieboyd: Does this logic let the Done button appear for statuses that aren't in the workflow sequence?  Opened for changes, etc

cases/services.py
BenAndrew-Kainos: Correct. They select which queues to be unassigned from

cases/views/main.py
BenAndrew-Kainos: Opened for changes does have a workflow sequence but yeah if it ends up at a status not in workflow it will currently show the button which can unassign from queues but wont ever increment status

BenAndrew-Kainos: Shabbir has confirmed this is OK

PR #412, opened 2020-03-18
LT-1595: f680 questions

a-bek:


cases/views/main.py
```diff
@@ -136,6 +138,45 @@ def post(self, request, **kwargs):
         )
 
 
+def get_additional_information(case):
+    field_titles = {
```
BenAndrew-Kainos: Could you leave a comment for what this is for (I'm not entirely sure)

cases/views/main.py
```diff
@@ -136,6 +138,45 @@ def post(self, request, **kwargs):
...
+                    "Description": title,
+                    "Answer": (
+                        friendly_boolean(value)
+                        if isinstance(value, bool)
+                        else value["value"]
+                        if isinstance(value, dict)
+                        else value
+                    ),
+                }
+            )
```
iichr: I like this approach, would've been quite impractical to add all of those titles and entries to the html 👌

PR #417, opened 2020-03-23
feat: LT-1060 - Non-UK organisation and sites

jfaracik:
* Added members view for organisations
* Consolidated organisation views
* Support for foreign addresses added


core/builtins/custom_tags.py
```diff
@@ -208,7 +208,7 @@ def default_na(value):
     """
     Returns N/A if the parameter given is none
     """
-    if value is not None:
+    if value:
```
andytaylorkainos: this will break if the value is a boolean

jfaracik: Changed it to `if value is not None and len(value):`

PR #431, opened 2020-03-27
fix: LT-2334 - Persist queue ID across case pages

jfaracik:
* Move queue cases page into queues app for separation between queues and cases
* Use reverse urls in cases page
* Cases page now uses built in pagination class


cases/urls.py
```diff
@@ -4,127 +4,113 @@
...
-    path("<uuid:pk>/team-advice-view/coalesce/", advice.CoalesceUserAdvice.as_view(), name="coalesce_user_advice"),
-    path("<uuid:pk>/team-advice-view/", advice.ViewTeamAdvice.as_view(), name="team_advice_view"),
-    path("<uuid:pk>/final-advice-view/coalesce/", advice.CoalesceTeamAdvice.as_view(), name="coalesce_team_advice"),
-    path("<uuid:pk>/final-advice-view/", advice.ViewFinalAdvice.as_view(), name="final_advice_view"),
-    path("<uuid:pk>/advice-view/give-user-advice/", advice.GiveUserAdvice.as_view(), name="give_user_advice"),
-    path(
-        "<uuid:pk>/advice-view/give-user-advice/<str:type>/",
+    path("", main.ViewCase.as_view(), name="case"),
+    path("done/", main.CaseProcessedByUser.as_view(), name="done"),
+    path("done/<uuid:queue_id>", main.CaseProcessedByUserForQueue.as_view(), name="done_for_queue"),
```
BenAndrew-Kainos: As the queue ID is now persistent we will need to merge these 2 views

conf/context_processors.py
```diff
@@ -7,9 +7,24 @@
...
+    if "queue_pk" not in request.resolver_match.kwargs:
+        return {}
+
+    queue_pk = request.resolver_match.kwargs["queue_pk"]
+    queue = get_queue(request, queue_pk)
+
+    print("\n")
+    print("queue")
+    print(queue)
+    print("\n")
```
BenAndrew-Kainos: Remember to remove in the final commit

BenAndrew-Kainos: Remember to remove in the final commit

conf/context_processors.py
```diff
@@ -7,9 +7,24 @@
...
+
+    return {"queue": {"id": queue_pk, "name": queue["name"], "is_system_queue": queue["is_system_queue"]}}
```
BenAndrew-Kainos: Is this necessary (can we not just. use get_queue which already returns this format)

conf/context_processors.py
```diff
@@ -23,15 +38,15 @@ def export_vars(request):
...
+            {"title": "Cases", "url": reverse_lazy("core:index"), "icon": "menu/cases"},
             {
                 "title": OrganisationsPage.TITLE,
                 "url": reverse_lazy("organisations:organisations"),
                 "icon": "menu/businesses",
             },
             {"title": "Teams", "url": reverse_lazy("teams:teams"), "icon": "menu/teams"},
             {"title": "My Team", "url": reverse_lazy("teams:team"), "icon": "menu/teams"},
-            {"title": QueuesList.TITLE, "url": reverse_lazy("queues:queues"), "icon": "menu/queues"},
+            # {"title": QueuesList.TITLE, "url": reverse_lazy("queues:queues"), "icon": "menu/queues"},
```
BenAndrew-Kainos: Why has this been commented out?

core/urls.py
```diff
@@ -5,6 +5,6 @@
 app_name = "core"
 
 urlpatterns = [
-    path("", views.hub, name="hub"),
+    path("", views.Index.as_view(), name="index"),
```
BenAndrew-Kainos: Nice :)

core/views.py
```diff
@@ -1,8 +1,10 @@
-from django.shortcuts import render, redirect
+from django.shortcuts import render
+from django.urls import reverse_lazy
+from django.views.generic import RedirectView
 
 
-def hub(request):
-    return redirect("/cases/")
+class Index(RedirectView):
+    url = reverse_lazy("queues:cases")
```
BenAndrew-Kainos: Should we not just put the index page class here

conf/context_processors.py
jfaracik: Done!

conf/context_processors.py
jfaracik: Done!

conf/context_processors.py
jfaracik: Uncommented again (was just for development)

core/views.py
jfaracik: Definitely an idea, I just wanted to keep queues grouped with queues, cases grouped with cases.

conf/context_processors.py
BenAndrew-Kainos: 👍

core/views.py
BenAndrew-Kainos: Sure that makes sense, can we just point the URL to that view then or does it need to be a redirect for something to do with queues?

jfaracik: Yeah can just point it :)

jfaracik: Done!

core/views.py
```diff
@@ -1,8 +1,4 @@
-from django.shortcuts import render, redirect
-
-
-def hub(request):
-    return redirect("/cases/")
+from django.shortcuts import render
```
BenAndrew-Kainos: Do we still need this file?

jfaracik: Sadly, used for the menu fallback for non-JS users.

PR #438, opened 2020-03-31
feat: LT-1258 - Flag colours

jfaracik:
image

**What's new**
* Users can now apply colours, labels and priorities to their flags

**Misc improvements**
* Added pagination to flags page (massive improvements in speeds and queries)
* Permissions service now uses `/me/` endpoint for better performance (half the time)
* Changed filtering to ONLY show deactivated flags on the flags page
* One consistent, accessible and functional include for showing flags on a page
* Service updates to match `/flags/` endpoint changes


flags/forms.py
```diff
@@ -33,20 +43,67 @@ def add_flag_form():
...
+                options=level_options,
             ),
+            RadioButtons(
+                title="Colour",
+                description="You can help convey information through use of colour",
+                name="colour",
+                classes=["app-radios--flag-colours"],
+                options=[
+                    Option("default", "Default"),
+                    Option("red", "Red", classes=["app-radios__item--red"]),
```
Mark-Spiers-kainos, line 55: would it be worth moving these options into a private variable in the file to call? means you don't have it in two places

flags/forms.py
```diff
@@ -33,20 +43,67 @@ def add_flag_form():
                 title=CreateFlagForm.Level.TITLE,
                 description=CreateFlagForm.Level.DESCRIPTION,
                 name="level",
-                options=options,
+                options=level_options,
             ),
+            RadioButtons(
+                title="Colour",
```
Mark-Spiers-kainos, line 49: lite-content

flags/views.py
```diff
@@ -80,30 +72,21 @@ class AddFlag(SingleFormView):
     def init(self, request, **kwargs):
         self.form = add_flag_form()
         self.action = post_flags
-
-    def get_success_url(self):
-        messages.success(self.request, flags.FlagsList.SUCCESS_MESSAGE)
-        return reverse("flags:flags")
+        self.data = {"colour": "default", "priority": 0}
+        self.success_message = flags.FlagsList.SUCCESS_MESSAGE
```
Mark-Spiers-kainos, line 76: could this be the cause of the banner on the cases page?

PR #440, opened 2020-04-02
feat: LT-1097 - HMRC SLA

andytaylorkainos:


core/builtins/custom_tags.py
```diff
@@ -330,20 +330,38 @@ def is_system_team(id: str):
 
 @register.filter()
 def get_sla_percentage(case):
-    if case["sla_remaining_days"] <= 0:
+    have_goods_departed = case.get("have_goods_departed")
```
BenAndrew-Kainos, line 333: This shouldn't be here, that should change whether SLA days is passed from the API

core/builtins/custom_tags.py
```diff
@@ -330,20 +330,38 @@ def is_system_team(id: str):
...
+        return "10"
+    else:
+        percentage = math.ceil(percentage / 10) * 10
+        return str(percentage)
 
 
 @register.filter()
-def get_sla_ring_colour(remaining_days):
+def get_sla_ring_colour(case):
+    have_goods_departed = case.get("have_goods_departed")
```
BenAndrew-Kainos, line 356: Same here

templates/queues/cases.html
```diff
@@ -92,9 +92,9 @@
 								</div>
 							</td>
 							<td class="cases-table__cell">
-								{% if case.sla_remaining_days != None %}
-									<div class="pie-wrapper progress-{{ case|get_sla_percentage }} {{ case.sla_remaining_days|get_sla_ring_colour }}" id="sla">
-										<span class="label">{{ case.sla_days }}</span>
+								{% if case.sla_remaining_days != None or case.sla_hours_since_raised != None %}
+									<div class="pie-wrapper progress-{{ case|get_sla_percentage }} {{ case|get_sla_ring_colour }}" id="sla">
+										<span class="label">{% if case.sla_remaining_days %}{{ case.sla_days }}{% else %}{{ case.sla_hours_since_raised }}{% endif %}</span>
```
BenAndrew-Kainos, line 97: Could you make the if statement to determine which function to use here, and then split the filter functions

core/builtins/custom_tags.py
andytaylorkainos: removed

core/builtins/custom_tags.py
andytaylorkainos: removed

core/builtins/custom_tags.py
```diff
@@ -330,20 +330,33 @@ def is_system_team(id: str):
 
 @register.filter()
 def get_sla_percentage(case):
-    if case["sla_remaining_days"] <= 0:
+    if case["case_type"]["sub_type"]["key"] == CaseType.HMRC:
```
BenAndrew-Kainos, line 333: Could you split this out too

andytaylorkainos: Done

PR #450, opened 2020-04-15
feat: LT-1331 - Trade Control Licences

andytaylorkainos:


templates/case/applications/open-licence-case.html
```diff
@@ -42,6 +42,18 @@
 				<p>{{ case.application.updated_at|str_date }}</p>
 			</div>
 		</div>
+		{% if case.case_type.reference.key == 'sicl' or case.case_type.reference.key == 'oicl' %}
```
Mark-Spiers-kainos, line 45: should this one not just not oicl? since it's in "open-licence-case.html"

templates/case/applications/standard-licence-case.html
```diff
@@ -42,6 +42,18 @@
 				<p>{{ case.application.updated_at|str_date }}</p>
 			</div>
 		</div>
+		{% if case.case_type.reference.key == 'sicl' or case.case_type.reference.key == 'oicl' %}
```
Mark-Spiers-kainos, line 45: likewise should this not just be "sicl"

templates/case/applications/open-licence-case.html
andytaylorkainos: Good spot!

templates/case/applications/standard-licence-case.html
andytaylorkainos: done!

PR #451, opened 2020-04-16
feat: LT-1021 Licensed Quantity & Licence page

BenAndrew-Kainos:


cases/views/advice.py
```diff
@@ -311,52 +312,61 @@ class Finalise(TemplateView):
     Finalise a case and change the case status to finalised
     """
 
+    @staticmethod
+    def _get_goods(request, pk, case_type):
+        goods = []
+        if case_type == CaseType.STANDARD.value:
+            goods, status_code = get_finalise_application_goods(request, pk)
+            if status_code != HTTPStatus.OK:
+                return error_page(request, "Approved goods could not be fetched")
```
andytaylorkainos, line 321: might be worth getting an Anthony-approved error message

cases/views/advice.py
```diff
@@ -372,12 +382,14 @@ def post(self, request, *args, **kwargs):
             return error_page(request, "You do not have permission.")
 
         if res.status_code == 400:
```
andytaylorkainos, line 384: i know its not your code but can you update this to the status library please?

cases/views/advice.py
BenAndrew-Kainos: will do

cases/views/advice.py
BenAndrew-Kainos: done

PR #452, opened 2020-04-16
feat: LT-2359 - Multiple control list entries

jfaracik:
**What's new**
* Now possible to add multiple control list entries throughout LITE
* Click a control list entry to view its details


cases/forms/review_goods_clc.py
```diff
@@ -19,23 +19,20 @@ def review_goods_clc_query_form(request, back_url, is_goods_type):
...
-                                inset_text=False,
                             ),
                         ],
                     ),
                     Option(key=conditional(is_goods_type, "False", "no"), value="No"),
                 ],
             ),
             RadioButtons(
-                title="Which report summary would you like to use? (optional)",
+                title="Which report summary would you like to use?",
```
BenAndrew-Kainos, line 32: Content for this page

cases/services.py
```diff
@@ -299,7 +303,7 @@ def get_good(request, pk):
 
 def get_goods_type(request, pk):
     data = get(request, GOODS_TYPE_URL + pk)
-    return data.json(), data.status_code
+    return {"good": data.json()}, data.status_code
```
BenAndrew-Kainos, line 306: Why are you doing this?

jfaracik:
```
API doesn't structure the endpoints in a way that flags (currently) works, so wrap data in dictionary
```

I've added a comment

cases/forms/review_goods_clc.py
jfaracik: Done

cases/services.py
BenAndrew-Kainos: I feel like we should fix that on the API. For now it's fine though

PR #453, opened 2020-04-21
bug: Picklists performance

BenAndrew-Kainos:
Load time for 5000 Picklists reduced from 7.56s to 0.03s (when paginated) or 0.27s (when not paginated for input lists)


templates/teams/picklists.html
```diff
@@ -72,6 +72,7 @@ <h4 class="govuk-heading-s">Deactivated</h4>
 					{% include "includes/information-text.html" with text="There aren't any items in this picklist" %}
 				{% endif %}
 			</div>
+			{% include "includes/pagination.html" with data=data %}
```
jfaracik, line 75: data=data isn't necessary

jfaracik: Also have the end to end tests been updated to handle pagination?

BenAndrew-Kainos: Removed data=data & the tests rely on the latest picklist being at the top (which it still is) so no changes required

templates/teams/picklists.html
```diff
@@ -72,6 +72,7 @@ <h4 class="govuk-heading-s">Deactivated</h4>
 					{% include "includes/information-text.html" with text="There aren't any items in this picklist" %}
 				{% endif %}
 			</div>
+			{% include "includes/pagination.html" %}
```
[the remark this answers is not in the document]
BenAndrew-Kainos: done :)

PR #455, opened 2020-04-22
bug: Queues performance

BenAndrew-Kainos:


queues/services.py
```diff
@@ -7,17 +7,16 @@
...
-    data = get(request, QUEUES_URL)
-    if convert_to_options:
-        converted = []
-
-        for queue in data.json().get("queues"):
-            converted.append(Option(queue.get("id"), queue.get("name"), description=queue.get("team").get("name")))
-
-        return converted
+def get_queues(request, disable_pagination=True, page=1, convert_to_options=False):
+    data = get(request, QUEUES_URL + "?page=" + str(page) + "&disable_pagination=" + str(disable_pagination)).json()
```
jfaracik, line 11: If you used an F string you wouldn't need to convert

templates/queues/manage.html
```diff
@@ -50,6 +50,7 @@ <h1 class="govuk-heading-l">
 				{% endfor %}
 			</tbody>
 		</table>
+		{% include "includes/pagination.html" with data=queues %}
```
jfaracik, line 53: {% pagination with data=queues %}

BenAndrew-Kainos: done

queues/services.py
BenAndrew-Kainos: done

PR #461, opened 2020-04-27
feat: LT-2310 - manage routing rules for all teams

Mark-Spiers-kainos:


assets/css/components/_routing-rules.scss
```diff
@@ -0,0 +1,4 @@
+.deactivated {
+	opacity: 0.65;
+	margin-bottom: - govuk-spacing(4);
+}
```
andytaylorkainos, line 4: new line pls

routing_rules/views.py
```diff
@@ -73,10 +73,20 @@ def get(self, request, **kwargs):
 
 class CreateRoutingRule(MultiFormView):
     def init(self, request, **kwargs):
+        select_team = has_permission(request, Permission.MANAGE_ALL_ROUTING_RULES)
         if request.method == "POST":
-            self.forms = routing_rule_form_group(request, request.POST.getlist("additional_rules[]"))
+            team_id = request.POST.get("team", get_gov_user(request)[0]["user"]["team"]["id"])
```
andytaylorkainos, line 78: can you make this a service, please?

assets/css/components/_routing-rules.scss
```diff
@@ -0,0 +1,4 @@
+.deactivated {
```
jfaracik, line 1: This should follow BEM (http://getbem.com) - Something like `.app-routing-rule__item--deactivated`

jfaracik: Also a new line at the end of the file :)

conf/context_processors.py
```diff
@@ -58,7 +58,8 @@ def lite_menu(request):
                 {"title": "Flagging rules", "url": reverse_lazy("flags:flagging_rules"), "icon": "menu/flags"},
             ),
             conditional(
-                has_permission(request, Permission.MANAGE_TEAM_ROUTING_RULES),
+                has_permission(request, Permission.MANAGE_TEAM_ROUTING_RULES)
```
jfaracik, line 61: We should be careful about calling has_permission multiple times in this file, I believe it does a request to the API for each one. Perhaps get all permissions on this page and then assert that the permission is in the list.

ui_automation_tests/pages/routing_rules_pages.py
```diff
@@ -19,6 +19,7 @@ class RoutingRulesPage(BasePage):
     SELECT_FLAG_ID = "flag"
     CHECKBOX_AND_LABEL = ".govuk-checkboxes__item"
     RADIO_BUTTONS = ".govuk-radios__input"
+    TEAM_PREFIX_ID = "team-"
```
jfaracik, line 22: TEAM_ID_PREFIX

templates/routing-rules/index.html
```diff
@@ -63,7 +63,11 @@ <h1 class="govuk-heading-l">
 						<td class="govuk-table__cell">
 							{% if rule.flags %}
 								{% for flag in rule.flags %}
-									{{ flag.name }}
+									{% if flag.status == "Active" %}
```
jfaracik, line 66: I'd be worried about doing an equal on a presentation value, does flag.status not have a key would compare to?

teams/services.py
```diff
@@ -45,3 +45,10 @@ def get_users_by_team(request, pk, convert_to_options=False):
     if convert_to_options:
         return [Option(user["id"], user["email"]) for user in data.json()["users"] if user["email"]]
     return data.json(), data.status_code
+
+
+def get_team_queues(request, pk, convert_to_option=False):
+    data = get(request, TEAMS_URL + pk + "/queues/")
+    if convert_to_option:
```
jfaracik, line 52: `convert_to_options`

And then only return `return data.json()`, we dont return status codes for gets

routing_rules/views.py
jfaracik: How would you make that into a service?

jfaracik: Could something like this work?

```
class CreateRoutingRule(MultiFormView):
        select_team = has_permission(request, Permission.MANAGE_ALL_ROUTING_RULES)
        team_id = request.POST.get("team", get_gov_user(request)[0]["user"]["team"]["id"])
        self.forms = routing_rule_form_group(
                request=request,
                additional_rules=request.POST.getlist("additional_rules[]", ()),
                team_id=team_id,
                select_team=select_team,
        )
        self.success_url = reverse("routing_rules:list")
        self.action = post_routing_rule
```

templates/routing-rules/index.html
Mark-Spiers-kainos: Flag status is returned as a string just

routing_rules/views.py
Mark-Spiers-kainos: That looks like it works well.

andytaylorkainos: > How would you make that into a service?

Woops - misread the line: assumed it was doing a POST

PR #462, opened 2020-04-28
feat: Approval blocking flags

BenAndrew-Kainos:


templates/case/advice/final.html
```diff
@@ -58,6 +58,16 @@ <h1 class="govuk-heading-l">
...
+		{% if blocking_flags %}
+			<div class="govuk-warning-text" id="blocking_flags">
+				<span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+				<strong class="govuk-warning-text__text">
+					<span class="govuk-warning-text__assistive">{% lcs 'cases.ViewAdvicePage.WarningBanner.HIDDEN_TEXT' %}</span>
+					{% lcs 'cases.ViewAdvicePage.WarningBanner.BLOCKING_FLAGS' %}{{ blocking_flags|join:"," }}
+				</strong>
+			</div>
+		{% endif %}
+
```
jfaracik, line 70: Flags are lacking border, colour and tooltip. Could you use the standard flags {% include %} for UI and code consistency. 

Screenshot 2020-04-29 at 15 00 16

BenAndrew-Kainos: But this is an error message. Would it have coloured flags in it?

flags/forms.py
```diff
@@ -102,6 +110,14 @@ def edit_flag_form():
             NumberInput(
                 name="priority", title=EditFlagForm.Priority.TITLE, description=EditFlagForm.Priority.DESCRIPTION
             ),
+            RadioButtons(
+                name="blocks_approval",
+                title=EditFlagForm.BlocksApproval.TITLE,
+                options=[
+                    Option("True", EditFlagForm.BlocksApproval.YES),
```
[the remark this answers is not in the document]
BenAndrew-Kainos: done

templates/case/advice/final.html
BenAndrew-Kainos: I will call design to find out what I should do

BenAndrew-Kainos: Design said it was better with colour so I've changed it

flags/views.py
```diff
@@ -81,6 +81,7 @@ class EditFlag(SingleFormView):
     def init(self, request, **kwargs):
         self.object_pk = str(kwargs["pk"])
         flag = get_flag(request, self.object_pk)
+        flag["blocks_approval"] = str(flag["blocks_approval"])
```
jfaracik, line 84: Is this still necessary?

BenAndrew-Kainos: let me test

BenAndrew-Kainos: no longer needed :)

PR #463, opened 2020-04-29
feat: LT-1413 - Default Queue

andytaylorkainos:


assets/javascripts/filter-default-queue-list.js
```diff
@@ -0,0 +1,20 @@
+let team = $("#team");
+
+function filterSelectOptions(team_value, default_to_select) {
+    // if team changes, set default_queue to 'select' and filter out other team queues
```
BenAndrew-Kainos, line 4: What is this used for?

core/helpers.py
```diff
@@ -91,3 +91,25 @@ def convert_value_to_query_param(key: str, value):
...
+    Returns None if a key is not found instead of throwing a KeyError
+    Example usage:
+    get_nested_value(case, ['application', 'end_user', 'name'])
+    """
+
+    value = dictionary.get(nested_keys.pop(0), None)
+
+    if value:
+        while nested_keys:
+            next_value = value.get(nested_keys.pop(0), None)
```
BenAndrew-Kainos, line 108: I'm very confused by this function. Does this not just return the last value that is None?

core/models.py
```diff
@@ -43,6 +45,7 @@ class User(AbstractUser):
     email = models.EmailField(default=None, blank=True, unique=True)
     first_name = models.CharField(max_length=30, blank=True, default=None, null=True)
     last_name = models.CharField(max_length=150, blank=True, default=None, null=True)
+    default_queue = models.UUIDField(default=uuid.UUID(ALL_CASES_QUEUE_ID), null=False)
```
BenAndrew-Kainos, line 48: Why is this needed?

queues/views.py
```diff
@@ -32,7 +32,7 @@ def get(self, request, **kwargs):
         case_type = request.GET.get("case_type")
         status = request.GET.get("status")
         sort = request.GET.get("sort")
-        queue_pk = kwargs["queue_pk"]
+        queue_pk = kwargs.get("queue_pk") or request.user.default_queue
```
BenAndrew-Kainos, line 35: Shouldn't this have a default if no kwarg or default queue

users/forms/users.py
```diff
@@ -24,8 +25,15 @@ def add_user_form(request):
...
             ),
+            Select(
+                title=AddUserForm.DefaultQueue.TITLE,
+                description=AddUserForm.DefaultQueue.DESCRIPTION,
+                name="default_queue",
+                options=get_queues(request, include_system=True, convert_to_options=True),
+            ),
         ],
         back_link=BackLink(AddUserForm.BACK_LINK, reverse_lazy("users:users")),
+        javascript_imports=["/assets/javascripts/filter-default-queue-list.js"],
```
BenAndrew-Kainos, line 36: I don't understand why this needs javascript

core/models.py
BenAndrew-Kainos: Nevermind, its for storing this locally right

core/helpers.py
andytaylorkainos: removed entirely

core/models.py
a-bek, line 48: How come we need to store this here + API?

andytaylorkainos: This is the way the frontends do "sessions"

PR #464, opened 2020-04-30
feat: LT-1230 - Open Licence Types (Media)

iichr:
- Hide the matrix for goods to countries when finalising

### DIFF
--- a/cases/views/advice.py
+++ b/cases/views/advice.py
@@ -326,7 +326,7 @@ class Finalise(TemplateView):
         case = get_case(request, str(kwargs["pk"]))
         case_type = case["application"]["case_type"]["sub_type"]["key"]
 
-        if case_type == CaseType.OPEN.value and (not case["application"]["goodstype_category"]["key"] == "media"):
+        if case_type == CaseType.OPEN.value and case["application"]["goodstype_category"]["key"] != "media":
             data = get_good_countries_decisions(request, str(kwargs["pk"]))["data"]
             items = [item["decision"]["key"] for item in data]
             is_open_licence = True
--- a/cases/views/advice.py
+++ b/cases/views/advice.py
@@ -326,7 +326,7 @@ class Finalise(TemplateView):
         case = get_case(request, str(kwargs["pk"]))
         case_type = case["application"]["case_type"]["sub_type"]["key"]
 
-        if case_type == CaseType.OPEN.value:
+        if case_type == CaseType.OPEN.value and (not case["application"]["goodstype_category"]["key"] == "media"):
             data = get_good_countries_decisions(request, str(kwargs["pk"]))["data"]
             items = [item["decision"]["key"] for item in data]
             is_open_licence = True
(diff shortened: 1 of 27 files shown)


templates/case/advice/final.html
```diff
@@ -39,9 +39,15 @@ <h1 class="govuk-heading-l">
...
 								{% if case.application.case_type.sub_type.key == 'open' %}
-									<a class="govuk-button"  href="{% url 'cases:finalise_goods_countries' queue.id case.id %}" id="finalise_button">
-										{% lcs 'cases.ViewAdvicePage.Actions.FINALISED_GOODS_AND_COUNTRIES' %}
-									</a>
+									{% if case.application.goodstype_category.key == 'media' %}
+										<a class="govuk-button"  href="{% url 'cases:finalise' queue.id case.id %}" id="finalise_button">
+										{% lcs 'cases.ViewAdvicePage.Actions.FINALISE' %}
+										</a>
+									{% else %}
+										<a class="govuk-button"  href="{% url 'cases:finalise_goods_countries' queue.id case.id %}" id="finalise_button">
```
jfaracik, line 47: Use a {% govuk_link_button %}

iichr: Copied that from just underneath, should we change both then?

iichr: Have a look changed it in a few places :)

cases/views/advice.py
```diff
@@ -326,7 +326,7 @@ def get(self, request, *args, **kwargs):
         case = get_case(request, str(kwargs["pk"]))
         case_type = case["application"]["case_type"]["sub_type"]["key"]
 
-        if case_type == CaseType.OPEN.value:
+        if case_type == CaseType.OPEN.value and (not case["application"]["goodstype_category"]["key"] == "media"):
```
a-bek, line 329: can you use `!=` instead of `not`?

iichr: On it

templates/case/advice/final.html
```diff
@@ -39,13 +39,16 @@ <h1 class="govuk-heading-l">
 							</button>
 							{% if can_advice_be_finalised %}
 								{% if case.application.case_type.sub_type.key == 'open' %}
-									<a class="govuk-button"  href="{% url 'cases:finalise_goods_countries' queue.id case.id %}" id="finalise_button">
-										{% lcs 'cases.ViewAdvicePage.Actions.FINALISED_GOODS_AND_COUNTRIES' %}
-									</a>
+									{% if case.application.goodstype_category.key == 'media' %}
+										{% make_list queue.id case.id as button_params %}
```
jfaracik, line 43: Could you just move the `{% make_list queue.id case.id as button_params %}` out of the if statement if its being reused?

Also pedantic, but IDs use dashes, not underscores :)

iichr: Done, changed to dashes as well

PR #475, opened 2020-05-07
feat: LT-2414 - Hawk Authentication

andytaylorkainos:


letter_templates/services.py
```diff
@@ -24,12 +24,12 @@ def get_letter_paragraphs(request, ids: list):
 
 
 def get_letter_template(request, pk, params=""):
-    data = get(request, LETTER_TEMPLATES_URL + pk + "?" + params)
+    data = get(request, LETTER_TEMPLATES_URL + pk + "/?" + params)
```
Mark-Spiers-kainos, line 27: could you just update the LETTER_TEMPLATES_URL to contain a "/" on the end?

PR #480, opened 2020-05-14
fix: Flakey tests

jfaracik:
🥇💯😎 69/69 😎💯🥇 

image


assets/css/components/_cases.scss
```diff
@@ -1,6 +1,6 @@
 .app-case__summary-list {
     display: grid;
-    grid-gap: govuk-spacing(3) govuk-spacing(6);
+    grid-gap: govuk-spacing(3) 150px;
```
BenAndrew-Kainos, line 3: Are we allowed to use px (I thought we had to use gov standards)

ui_automation_tests/pages/assign_flags_to_case.py
```diff
@@ -1,8 +1,10 @@
+import time
 from shared import selectors
 from shared.BasePage import BasePage
 
 
 class CaseFlagsPages(BasePage):
     def select_flag(self, flag_name):
+        time.sleep(0.5)
```
BenAndrew-Kainos, line 8: Once again please comment why a sleep is required

ui_automation_tests/pages/case_page.py
```diff
@@ -41,6 +41,8 @@ def click_change_case_flags(self):
         self.driver.find_element_by_id(self.LINK_CHANGE_CASE_FLAGS_ID).click()
 
     def click_assign_case_officer(self):
+        Shared(self.driver).set_header_to_not_stick()
```
BenAndrew-Kainos, line 44: What is this for?

assets/css/components/_cases.scss
jfaracik: `govuk-spacing` is just a helper function for common spacing units. It's absolutely fine to use custom sizes for two reasons: They don't have a spacing unit for this, and that this is also internal.

ui_automation_tests/pages/case_page.py
phlptt: It basically sets the header to stay to the top of the page rather than stay on the page whilst you scroll which is making the tests fail

ui_automation_tests/pages/assign_flags_to_case.py
phlptt: temporary fix to make the tests pass. i'll add in a todo to make it an implicit wait

BenAndrew-Kainos: Cool. So long as it's replaced at some point

ui_automation_tests/pages/case_page.py
BenAndrew-Kainos: As discussed please investigate and remove or comment

PR #483, opened 2020-05-14
fix: Show end use details and route of goods on standard cases

jfaracik:
* And also removes end use details for crypto OIELs!

### DIFF
--- a/templates/case/slices/end-use-details.html
+++ b/templates/case/slices/end-use-details.html
@@ -26,8 +26,10 @@
 				{% lcs 'cases.ApplicationPage.EndUseDetails.INFORMED_TO_APPLY_TITLE' %}
 			</td>
 			<td class="govuk-table__cell">
-				<p class="govuk-body govuk-!-margin-bottom-1">{{ case.application.is_military_end_use_controls|friendly_boolean }}</p>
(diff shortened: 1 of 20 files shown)


templates/case/slices/end-use-details.html
```diff
@@ -18,41 +18,76 @@
...
-			</td>
-			<td class="govuk-table__cell govuk-!-font-weight-bold">
-				{% lcs 'cases.ApplicationPage.EndUseDetails.SUSPECTED_WMD_TITLE' %}
-			</td>
-			<td class="govuk-table__cell">
-				<p class="govuk-body govuk-!-margin-bottom-1">{{ case.application.is_suspected_wmd|friendly_boolean }}</p>
-				<p class="govuk-hint govuk-!-margin-0" data-max-length="200">{{ case.application.suspected_wmd_ref }}</p>
-			</td>
-		</tr>
+		{% if case.sub_type == "standard" or case.sub_type == "open" %}
```
iichr, line 21: not all open applications have this section e.g the `cryptographic` OIEL

jfaracik: Great spot! Fixed :) 🚀

PR #491, opened 2020-05-18
fix: Show Status + ARS 

jfaracik:


templates/case/slices/goods.html
```diff
@@ -138,6 +162,18 @@ <h2 class="govuk-heading-m">{{ case.goods|length }} good{{ case.goods|pluralize
 								{% endfor %}
 							</span>
 						</td>
+						{% if not hide_status %}
+						{{ good }}
```
iichr, line 166: indentation /not blocking :)/

jfaracik: Good spot! Shouldn't have been there :P

PR #492, opened 2020-05-18
fix: OIEL improvements

jfaracik:


cases/views/main.py
```diff
@@ -53,9 +53,9 @@ def get_open_application(self):
             Slices.DESTINATIONS,
             conditional(self.case.data["inactive_parties"], Slices.DELETED_ENTITIES),
             Slices.LOCATIONS,
-            conditional(self.case.data["goodstype_category"]["key"] != "cryptographic", Slices.END_USE_DETAILS),
-            Slices.ROUTE_OF_GOODS,
+            conditional(self.case.data["goodstype_category"]["key"] != "cryptographic", *[Slices.END_USE_DETAILS, Slices.ROUTE_OF_GOODS]),
```
iichr: Crypto's a weird one in that it doesn't have `end use details`. As for route of goods, we automatically put those down as a no (technically not a section the user can fill in)

iichr: Disregards, just looked again and seen the negation 🤦 _this never happened_

PR #495, opened 2020-05-19
enhance: hawk response nonce checking

efahy:
…gging warnings


BenAndrew-Kainos:
Was your change to lite_forms intentional? (I don't see a PR for it)

efahy:
> Was your change to lite_forms intentional? (I don't see a PR for it)

I just pulled in the latest master - you think that's bad?

PR #496, opened 2020-05-20
feat: LT-2146 - OIEL and OITL Destination Contract Types

iichr:
- Destination and accompanying contract type flags are now returned by the api in a different format - matched on the frontend


templates/case/pills/destinations.html
```diff
@@ -8,7 +8,7 @@
 				<p class="govuk-!-margin-0">{{ destination.name }} <span class="govuk-hint govuk-!-margin-0 lite-!-display-inline">({{ destination.type|sentence_case }})</span></p>
 				<p class="govuk-hint govuk-!-margin-0">{{ destination|get_address }}</p>
```
iichr, line 9: @jfaracik There are just for standard right?

jfaracik: Yup! Looks good to me

PR #497, opened 2020-05-20
spike: Universal document context

BenAndrew-Kainos:


BenAndrew-Kainos:
> Is there a need for this PR? seems a bit weird it just adds an image

Can't add it otherwise :)

PR #499, opened 2020-05-22
fix: Remove ECJU query types

jfaracik:
* Removes two types of ECJU queries for licence applications, will be added back for compliance cases in a future story


ui_automation_tests/features/ecju_queries.feature
```diff
@@ -12,7 +12,6 @@ Feature: I want to create ECJU queries
     When I go to application previously created
     And I go to the ECJU queries tab
     And I click new query
-    And I select standard ECJU Query for my query type
```
BenAndrew-Kainos, line 15: Why is this no longer needed?

BenAndrew-Kainos, line 15: Please leave a comment to say this will be added in with that later story (include story number)

jfaracik: It’s going to have a different flow so that step won’t be added back anyway.

PR #503, opened 2020-05-27
feat: 1310 Import Enforcement XML

BenAndrew-Kainos:


queues/services.py
```diff
@@ -83,3 +84,18 @@ def get_enforcement_xml(request, queue_pk):
         return response, data.status_code
     else:
         return None, data.status_code
+
+
+def post_enforcement_xml(request, queue_pk, json):
+    if len(request.FILES) == 0:
```
jfaracik, line 90: Could it be possible that request.FILES could return None as opposed to a list?

templates/queues/cases.html
```diff
@@ -28,6 +28,7 @@
 				</div>
 				{% if enforcement_check and data.results.cases %}
 					<a id="export-xml" class="govuk-button govuk-button--secondary" href="{% url 'queues:enforcement_xml_export' queue.id %}" draggable="false" role="button">{% lcs 'cases.CasesListPage.EnforcementXML.EXPORT_BUTTON' %}</a>
+					<a id="export-xml" class="govuk-button govuk-button--secondary" href="{% url 'queues:enforcement_xml_import' queue.id %}" draggable="false" role="button">{% lcs 'cases.CasesListPage.EnforcementXML.IMPORT_BUTTON' %}</a>
```
jfaracik, line 31: Would you mind using {% govuk_link_button %} instead? It would be something along the lines of `{% govuk_link_button text='cases.CasesListPage.EnforcementXML.IMPORT_BUTTON' id='export-xml' url='queues:enforcement_xml_import' params=queue.id %}`

queues/services.py
BenAndrew-Kainos: No files is always a list. I'll change to `if request.FILES` anyway because it looks nice

BenAndrew-Kainos: nevermind, doesn't work if you do if request.FILES

templates/queues/cases.html
BenAndrew-Kainos: There doesn't seem to be a params arg?

BenAndrew-Kainos: done

BenAndrew-Kainos: done

PR #506, opened 2020-05-28
feat: LT-1551 - Filter picklists

jfaracik:
**What's new**
* Now possible to filter picklist items

**Misc**
* Updated GOV.UK frontend to v3.7 (allows us to remove our custom back link styling ❤️)
* Improved picklists UI
* Removed redirect if picklist type isn't provided
* Moved permission to manage picklists to app level 
* Fixed editing picklist item bug where existing text doesn't show
* Fixed `lite-header` changing size depending on the page


conf/urls.py
```diff
@@ -11,7 +11,7 @@
     path("open-general-licences/", include("open_general_licences.urls")),
     path("organisations/", include("organisations.urls")),
     path("queues/", include("queues.urls")),
-    path("team/picklists/", include("picklists.urls")),
```
iichr, line 14: 👀

iichr: sorry should've been the next line, might need the `/` I remember for HAWK

jfaracik: I've changed it back :)

PR #508, opened 2020-06-01
display site audit trail

stephanieboyd:


conf/constants.py
```diff
@@ -26,6 +26,7 @@ class GoodSystemFlags:
 FINAL_ADVICE_URL = "/final-advice/"
 VIEW_FINAL_ADVICE_URL = "/view-final-advice/"
 ACTIVITY_URL = "/activity/"
+SITE_ACTIVITY_URL = "/sites-activity/"
```
efahy, line 29: Maybe the constant name could reflect that this is a continuation of an organisations URL to avoid the possibility of someone trying to use it as a root URL?  Something like `ORGANISATION_SITES_ACTIVITY_URL` maybe?

iichr: 👍 Addressed!

PR #510, opened 2020-06-02
feat: 1252 Set document addressee 

BenAndrew-Kainos:
Update generate document views & adding addressee step


cases/services.py
```diff
@@ -324,9 +324,8 @@ def post_generated_document(request, pk, json):
     return data.status_code
 
 
-def get_generated_document_preview(request, pk, tpk, text):
-    text = quote(text)
-    data = get(request, CASE_URL + pk + GENERATED_DOCUMENTS_PREVIEW_URL + "?template=" + tpk + "&text=" + text)
+def get_generated_document_preview(request, pk, params):
```
jfaracik, line 327: Could you make these params explicit?

BenAndrew-Kainos: I thought we had switched to making them non-explicit?

BenAndrew-Kainos: Done

cases/forms/generate_document.py
```diff
@@ -1,41 +1,44 @@
...
+def select_template_form(templates, back_url):
     return Form(
         title=letter_templates.LetterTemplatesPage.PickTemplate.TITLE,
         questions=[
             RadioButtonsImage(
                 name="template",
                 options=[
                     Option(t["id"], t["name"], img_url=f"/assets/images/letter_templates/{t['layout']['filename']}.png")
-                    for t in templates
+                    for t in templates["results"]
```
rorysproule1, line 16: Possible to add a clearer value than t?

BenAndrew-Kainos: Good suggestion. Done

PR #521, opened 2020-06-05
fix: LT-2713 - Deactivate and reactivate picklist items

jfaracik:


open_general_licences/views.py
```diff
@@ -136,14 +136,15 @@ def init(self, request, **kwargs):
             if self.kwargs["status"] == "reactivate"
             else open_general_licences_strings.Deactivate
         )
-        self.success_message = self.strings.SUCCESS_MESSAGE
+        if request.POST.get("response") == "yes":
```
a-bek, line 139: use str_to_bool?

jfaracik: No such thing on internal ⭕

PR #522, opened 2020-06-05
task: Add new filters

BenAndrew-Kainos:


letter_templates/views/view.py
```diff
@@ -6,13 +6,23 @@
...
+        name = request.GET.get("name")
+        if name:
+            params["name"] = name
+
         data, _ = get_letter_templates(request, convert_dict_to_query_params(params))
-        context = {"data": data, "page": params.pop("page")}
+        filters = FiltersBar([TextInput(name="name", title="name"),])
+        context = {
+            "data": data,
+            "page": params.pop("page"),
```
jfaracik, line 23: Do we still need page in context?

jfaracik: + could params not be `params = {"page": int(request.GET.get("page", 1), "name": ...)}`

users/views/users.py
```diff
@@ -23,9 +23,12 @@
 class UsersList(TemplateView):
     def get(self, request, **kwargs):
         status = self.request.GET.get("status")
+        email = self.request.GET.get("email")
         params = {"page": int(self.request.GET.get("page", 1))}
```
jfaracik, line 27: Same comment as above.

BenAndrew-Kainos: done

letter_templates/views/view.py
BenAndrew-Kainos: done

letter_templates/views/view.py
```diff
@@ -6,13 +6,19 @@
...
 )
+from lite_forms.components import FiltersBar, TextInput
 
 
 class LetterTemplatesList(TemplateView):
     def get(self, request, **kwargs):
-        params = {"page": int(request.GET.get("page", 1))}
+        params = {"page": int(request.GET.get("page", 1)), "name": request.GET.get("name", "")}
+
         data, _ = get_letter_templates(request, convert_dict_to_query_params(params))
```
jfaracik, line 16: Use convert_parameters_to_query_params instead of convert_dict_to_query_params :)

jfaracik: Brownie points for doing `convert_parameters_to_query_params` inside of `get_letter_templates` instead

BenAndrew-Kainos: done

PR #530, opened 2020-06-11
feat: LT-2110 - Register for OGLs

jfaracik:
* Added support for OGL cases
* Cleaned up references to case application/query data, now just `case.data` globally
* Enabled GZIP compression


cases/constants.py
```diff
@@ -3,6 +3,8 @@
 
 class CaseType(Enum):
     APPLICATION = "application"
+    QUERY = "query"
+    REGISTRATION = "registration"
```
BenAndrew-Kainos, line 7: Could you rename to OGL registration

core/services.py
```diff
@@ -88,25 +88,39 @@ def get_permissible_statuses(request, case):
...
                     goods_query_status_keys.insert(1, Statuses.PV)
 
             case_type_applicable_statuses = [status for status in statuses if status["key"] in goods_query_status_keys]
-
+    elif case_type == CaseType.REGISTRATION.value:
+        case_type_applicable_statuses = [
+            status
+            for status in statuses
+            if status["key"]
+            in [
```
BenAndrew-Kainos, line 114: Could you move this list to another constant please

templates/case/slices/open-general-licence.html
```diff
@@ -0,0 +1,69 @@
+<h2 class="govuk-heading-m">Site</h2>
+<dl class="app-case__summary-list">
+	<div class="app-case__summary-list__row">
+		<dt class="app-case__summary-list__key">
+			Name
```
BenAndrew-Kainos, line 5: Content (multiple in this file)

templates/case/slices/open-general-licence.html
```diff
@@ -0,0 +1,69 @@
...
+		<dt class="app-case__summary-list__key">
+			More information
+		</dt>
+		<dd class="app-case__summary-list__value">
+			{{ case.data.open_general_licence.url|linkify }}
+		</dd>
+	</div>
+</dl>
+
+<br><br><br>
```
BenAndrew-Kainos, line 67: This looks suspect

cases/constants.py
jfaracik: Registration is a new type of case, in the same way application and query are.

core/services.py
```diff
@@ -88,25 +88,39 @@ def get_permissible_statuses(request, case):
...
+                Statuses.FINALISED,
+                Statuses.REGISTERED,
```
Mark-Spiers-kainos, line 116: should FINALISED be first in the list? seems a bit weird, since finalised means licence is already created and documents generated.

core/services.py
jfaracik: Seeing as it's only used the once and it follows the format of the statuses previous, is there any point?

core/services.py
jfaracik: I'm not too sure to be honest, we'd have to see what research suggests.

core/services.py
BenAndrew-Kainos: for configurations it's best if we can control all the available statuses in one place

ui_automation_tests/pages/shared.py
```diff
@@ -102,3 +103,6 @@ def filter_by_name(self, name):
         self.driver.find_element_by_id(self.INPUT_NAME_FILTER_ID).clear()
         self.driver.find_element_by_id(self.INPUT_NAME_FILTER_ID).send_keys(name)
         self.driver.find_element_by_css_selector(selectors.BUTTON_APPLY_FILTERS).click()
+
+    def click_first_link_in_row(self):
+        self.driver.find_element_by_css_selector(self.FIRST_LINK_IN_ROW).click()
```
BenAndrew-Kainos, line 108: please could you give the link an ID instead rather than relying on it being first

ui_automation_tests/step_defs/test_open_general_licences.py
```diff
@@ -114,3 +116,33 @@ def see_the_newly_generated_open_general_export_licence(driver, context):
...
+@when("I click on first case")
+def click_on_first_case(driver):
+    Shared(driver).click_first_link_in_row()
+
+
+@then("I see OGEL case")
+def see_ogel(driver):
+    ogel = OpenGeneralLicencesCasePage(driver)
+    assert "Site" in ogel.get_text_of_first_heading()
+    assert "Open General Export Licence" in ogel.get_text_of_second_heading()
```
BenAndrew-Kainos, line 148: Please do not assert content. Instead assert that expected properties are visible on the case page

ui_automation_tests/pages/open_general_licences_pages.py
```diff
@@ -75,3 +75,14 @@ class OpenGeneralLicencesDeactivatePage(BasePage):
 
     def select_yes(self):
         self.driver.find_element_by_id(self.RADIO_YES_ID).click()
+
+
+class OpenGeneralLicencesCasePage(BasePage):
+    HEADING = ".govuk-heading-m"
+
+    def get_text_of_first_heading(self):
+        return self.driver.find_elements_by_css_selector(self.HEADING)[0].text
```
BenAndrew-Kainos, line 84: We shouldn't be checking headings

core/services.py
BenAndrew-Kainos: Is that research? Surely that's criteria of the ticket?

core/services.py
jfaracik: Agreed, that place should be in the API and that should have been raised in the initial ticket for statuses.

core/services.py
jfaracik: I mean it's ordering, it isn't too much of an issue.

PR #535, opened 2020-06-16
feat: 2642 document variable help

BenAndrew-Kainos:


core/builtins/custom_tags.py
```diff
@@ -596,6 +596,11 @@ def sentence_case(text):
     return capfirst(text).replace("_", " ")
 
 
+@register.filter()
+def format_heading(text):
+    return text.replace("_", " ").capitalize()
```
efahy, line 601: As discussed, remove `capitalize` and make the enums the correct case, thus allowing ECJU to be all uppercase, for example

letter_templates/views/create.py
```diff
@@ -48,3 +49,18 @@ def post(request):
             return error_page(None, "; ".join(error_messages))
 
         return redirect("letter_templates:letter_templates")
+
+
+class VariableHelp(TemplateView):
+    @staticmethod
```
efahy, line 55: I think a docstring comment here would be good, explaining the general concept of the business configureable enums and text being dynamically turned into a rendered help page

BenAndrew-Kainos: Done

core/builtins/custom_tags.py
BenAndrew-Kainos: done

PR #537, opened 2020-06-18
feat: LT-2724-view-open-licence-returns

rorysproule1:


compliance/services.py
```diff
@@ -0,0 +1,19 @@
+from django.http import HttpResponse
+
+from conf.client import get
+from conf.constants import OPEN_LICENCE_RETURNS_URL, COMPLIANCE_URL, COMPLIANCE_LICENCES_URL
+
+FILENAME = "OpenLicenceReturns.csv"
+
+
+def get_compliance_licences(request, case_id, reference, page):
+    data = get(request, COMPLIANCE_URL + case_id + COMPLIANCE_LICENCES_URL + f"?reference={reference}&page={page}",)
```
andytaylorkainos, line 10: if the reference is none, will it not try to filter `reference=None&page=1`?

rorysproule1: The usage for this function provides a defaulted empty string if the reference isnt found

PR #543, opened 2020-06-22
bug: 1379 document upload

BenAndrew-Kainos:


cases/helpers/documents.py
```diff
@@ -0,0 +1,67 @@
...
+        return _extract_document_data(files, request.POST.get("description")), None
+    except UploadFailed:
+        if request.upload_handlers[0].content_length > MAX_UPLOAD_SIZE:
+            return None, cases.Manage.Documents.AttachDocuments.FILE_TOO_LARGE
+        else:
+            return None, cases.Manage.Documents.AttachDocuments.UPLOAD_FAILURE_ERROR
+
+
+def generate_file(result):
+    # Stream file upload
```
efahy, line 54: This function is called from `download_document_from_s3` so shouldn't this comment be "Stream file download"

cases/views/main.py
```diff
@@ -268,38 +265,30 @@ def init(self, request, **kwargs):
...
+        if error:
+            return form_page(
+                request, self._get_form(kwargs), extra_data={"case_id": case_id, "case": case}, errors={"file": [error]}
+            )
 
         # Send LITE API the file information
-        case_documents, _ = post_case_documents(request, case_id, data)
+        case_documents, _ = post_case_documents(request, case_id, [data])
 
         if "errors" in case_documents:
```
efahy, line 293: If the post to the API fails, it leaves an orphaned file in S3, does it not?  We should remove it

PR #553, opened 2020-06-26
feat: LT-1410 - Reissue Licence

a-bek:


templates/case/tabs/licences.html
```diff
@@ -0,0 +1,36 @@
+{% if case.licences %}
+	<table id="table-licences" class="govuk-table">
+		<thead class="govuk-table__head">
+			<tr class="govuk-table__row">
+				<th class="govuk-table__header" scope="col">Licence</th>
```
andytaylorkainos, line 5: content for headers

BenAndrew-Kainos: done

PR #564, opened 2020-07-07
feat: LT-1446-outstanding-annual-return

rorysproule1:


templates/case/tabs/compliance-licences.html
```diff
@@ -20,6 +23,23 @@
 					<td class="govuk-table__cell govuk-table__cell--tight">
 						{{ licence.status.value }}
 					</td>
+					<td class="govuk-table__cell govuk-table__cell--tight">
+						{% if licence.case_type.sub_type.key == "open" %}
```
Mark-Spiers-kainos, line 27: can standard application cases licences not have open licence returns? As far as I'm aware it should be any licence

rorysproule1: As discussed, OLRs are relevant to open licences

PR #565, opened 2020-07-07
feat: LT-1188 - I want to have the option to add a note whenever I progress a case

jfaracik:
**What's new**
* Added optional note fields for certain actions, such as changing status or assigning users to cases

**Misc**
* _I'm Done_ now extends SingleFormView, errors are handled inline in the form now :)
* _Assign users (queue edition)_ now extends SingleFormView, has a friendly message on success :)

### DIFF
--- a/users/services.py
+++ b/users/services.py
@@ -34,6 +34,7 @@ def get_gov_users(request, params=None, convert_to_options=False):
                 value = email
                 description = None
 
+            # Hide users without emails (eg system users)
             if email:
                 converted.append(Option(key=user.get("id"), value=value, description=description))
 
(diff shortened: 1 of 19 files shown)


users/services.py
```diff
@@ -34,7 +34,9 @@ def get_gov_users(request, params=None, convert_to_options=False):
                 value = email
                 description = None
 
-            converted.append(Option(key=user.get("id"), value=value, description=description))
+            # Hide users without emails (eg system users)
+            if email:
```
Mark-Spiers-kainos, line 38: I think this condition should be done before the previous if else, seems weird to set value/description based on email then do nothing with it.

jfaracik: Done 🚀

PR #574, opened 2020-07-15
bugs: document templates - 2764, 2805, 2681, 2696, 2847

rorysproule1:


templates/letter-templates/add-letter-paragraphs.html
```diff
@@ -46,6 +46,6 @@ <h1 class="govuk-fieldset__heading">
 			</button>
 		</form>
 	{% else %}
-		{% include "includes/notice.html" with text="There aren't any letter paragraphs left to add." %}
+		{% include "includes/notice.html" with text='LetterTemplates.AddParagraph.NO_LETTER_PARAGRAPHS' %}
```
iichr, line 49: does that work when it's wrapped in ' ', would've thought it would print the path as a string ?

rorysproule1: as discussed, works with both!

PR #588, opened 2020-07-22
bug: OIEL finalising

BenAndrew-Kainos:
James pointed out that all open application types should use the good/country matrix now, so removing old code blocking that


BenAndrew-Kainos:
James pointed out that all open application types should use the good/country matrix now, so removing old code blocking that

PR #595, opened 2020-08-03
Fix view case

richtier:


.circleci/config.yml
```diff
@@ -58,9 +62,13 @@ jobs:
           key: dependencies-{{ .Branch }}-{{ checksum "Pipfile.lock" }}
 
       - run:
-          name: Run tests
+          name: Run testcase
           command: pipenv run ./manage.py test --parallel
 
+      - run:
```
r4vi, line 68: what's the reason for running `./manage.py test --parallel` and running `pytest`?

richtier: `manage.py test` runs django tests (TestCase tests)

pytest is pytest

cards on the table: when I started writing the pytest test I did not know there were TestCase tests.

unfortunately pytest fixtures do not work with TestCase tests

but rewriting the TestCase tests to work in pytest wont take much effort - they're rather simple.

